PR #707, opened 2026-02-27
feat: approval gate MCP proxy with Authentik SSO and port isolation

agentydragon-agent:
## Summary

- **Approval gate**: a generic MCP proxy that wraps any backend MCP server with a human-in-the-loop approval layer. Agent tool calls are queued; on operator approval the gate forwards the call to the backend and delivers results via MCP `ResourceUpdated` notifications → OpenClaw `chat.inject`.
- **Authentik SSO** for the operator UI: admins sign in via Authentik; no manual token entry. The UI trusts `X-Authentik-Username`/`X-Authentik-Groups` headers injected by the outpost.
- **Port isolation** (CiliumNetworkPolicy) prevents OpenClaw from forging operator approvals: port 8765 (UI+API) is only reachable from the `authentik` namespace; port 8767 (MCP) is only reachable from `openclaw`.

## Components

**`approval_gate/`** (new package):
- `models.py` — discriminated union state machine: `Pending → Executing → Done/Rejected/Withdrawn`
- `storage.py` — aiosqlite action store
- `predicates.py` — three-way auto-approval: `Approved | Denied | NeedsHumanDecision`
- `proxy_server.py` — FastMCP proxy: discovers backend tools, wraps schemas, emits `ResourceUpdated`
- `operator_api.py` — REST API (OPERATOR_API_KEY bearer)
- `ui.py` — Jinja2 operator UI (Authentik header auth)
- `app.py` — `create_apps()` returns `(operator_app, mcp_asgi)` for dual-port serving
- `main.py` — `asyncio.gather()` of two `uvicorn.Server` instances

**`openclaw/approval-gate/`** (new): TypeScript plugin subscribing to `resource://actions/{id}` notifications, translating to `chat.inject`.

**`mcp_infra/exec/direct_server.py`**: standalone exec backend entry point for the exec container.

**k8s** (`cluster/k8s/approval-gate/`): namespace, ExternalSecret (Vault), PVC, deployments, services, Authentik HTTPRoute, `CiliumNetworkPolicy`.

**Authentik** (`cluster/k8s/authentik/blueprints/`): `approval-gate-sso.yaml` (proxy provider + admins-only policy binding), `shared-proxy-outpost.yaml` updated.

**Docker**: `docker/approval-gate/`, `docker/approval-gate-exec-backend/`, `docker/openclaw/Dockerfile` updated.

## Test plan

- [ ] `bazel test //approval_gate/...` — 3 tests pass (models, storage, predicates)
- [ ] Deploy to cluster; visit `https://approval-gate.allegedly.works` — Authentik login, then action queue
- [ ] Call a tool via MCP client with AGENT_API_KEY — action appears in UI
- [ ] Approve — backend executes, ResourceUpdated emitted
- [ ] Verify OpenClaw plugin receives notification and injects result into chat
- [ ] Confirm OpenClaw pod cannot reach port 8765 (CNP blocks it)

https://claude.ai/code/session_01FvbBvDb846eqQq6qaQdTwH


approval_gate/proxy_server.py
```diff
@@ -0,0 +1,233 @@
...
+            decision = call_predicate(self._predicate, tool_name, input)
+            match decision:
+                case Approved():
+                    self._spawn(self.decide(action_id, ApproveDecision()))
+                case Denied(reason=reason):
+                    self._spawn(self.decide(action_id, DenyDecision(reason=reason or "automatically denied")))
+                case NeedsHumanDecision():
+                    logger.info("queued action id=%s tool=%s", action_id, tool_name)
+
+            return response.model_dump()
```
agentydragon: just return pydantic, fastmcp can handle it

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,233 @@
...
+        ) -> dict[str, object]:
+            call = ToolCall(tool_name=tool_name, arguments=input)
+            action_id = str(uuid.uuid4())
+            await self._storage.create(
+                action_id=action_id, call=call, justification=justification, session_key=session_key
+            )
+            await self.broadcast_resource_list_changed()
+            response = ActionRef(action_id=action_id)
+
+            decision = call_predicate(self._predicate, tool_name, input)
```
agentydragon: the response should be returned immediately, before even calling the predicate. the predicate should be called asynchronously outside of the tool's execution.

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,233 @@
...
+        Raises ValueError if the action does not exist or is not pending.
+        """
+        action = _require_action(await self._storage.get(action_id), action_id)
+        if not isinstance(action.state, PendingState):
+            raise ValueError(f"Action {action_id} is not pending (status={action.state.status!r})")
+
+        match decision:
+            case ApproveDecision():
+                action = await self._update_and_notify(action_id, ExecutingState())
+                outcome = await self._execute_backend_call(action)
```
agentydragon: the "allow" decision should return approx immediately. it should not block until it's done executing. i expect you'll need some kind of background event loop or something to handle things like this.

approval_gate/storage.py
```diff
@@ -0,0 +1,137 @@
...
+        state=_ACTION_STATE_TA.validate_json(row.state_json),
+    )
+
+
+class ActionStorage:
+    """Async SQLite storage for action records."""
+
+    def __init__(self, db_path: Path) -> None:
+        self._db_path = db_path
+        self._session_factory: async_sessionmaker[AsyncSession] | None = None
```
agentydragon: this should not be None, ctor should require it and initialize whould be a factory method

approval_gate/storage.py
```diff
@@ -0,0 +1,137 @@
...
+        """List actions, optionally filtered by status, newest first."""
+        async with self._sessions() as session:
+            stmt = select(_ActionRow).order_by(_ActionRow.created_at.desc()).limit(limit)
+            if status is not None:
+                stmt = stmt.where(_ActionRow.status == status)
+            result = await session.execute(stmt)
+            rows = result.scalars().all()
+        return [_row_to_action(r) for r in rows]
+
+    async def list_pending(self) -> list[Action]:
```
agentydragon: just inline list_by_status wherever this is called

approval_gate/test_storage.py
```diff
@@ -0,0 +1,88 @@
...
+
+from __future__ import annotations
+
+import pytest_bazel
+from mcp.types import CallToolResult, TextContent
+
+from approval_gate.models import ActionStatus, DoneState, PendingState, RejectedState, ToolCall
+from approval_gate.storage import ActionStorage
+
+# storage fixture is provided by conftest.py
```
agentydragon: not useful comment

openclaw/approval-gate/index.ts
```diff
@@ -0,0 +1,402 @@
...
+
+type PendingInject = { resolve: () => void; reject: (err: Error) => void; timeout: NodeJS.Timeout };
+
+/**
+ * Persistent WebSocket connection to the OpenClaw gateway.
+ * Authenticates once on connect and reuses the socket for all chat.inject calls.
+ * Automatically reconnects on close; queues inject calls that arrive before auth completes.
+ */
+class GatewayConnection {
+  private ws: WebSocket | null = null;
```
agentydragon, line 133: i would expect that openclaw would provide some kind of class to handle connecting to gateway? check how their plugin-writing boilerplate looks in openclaw repo

openclaw/approval-gate/index.ts
```diff
@@ -0,0 +1,402 @@
...
+          }
+          return { content: [{ type: "text" as const, text: formatOutcomeMessage(action) }] };
+        }
+
+        // Still pending — tell the agent its action is queued.
+        return {
+          content: [
+            {
+              type: "text" as const,
+              text: `Action ${actionId.slice(0, 8)} queued for operator approval${approvalUrl ? ` at ${approvalUrl}` : ""}`,
```
agentydragon: don't truncate action ids for the agent - it needs to know the full action id so it can reference it later

openclaw/approval-gate/index.ts
```diff
@@ -0,0 +1,402 @@
...
+            },
+          ],
+        };
+      },
+    }));
+  }
+
+  api.logger.info(
+    `approval-gate: registered ${toolList.tools.length} tool(s): ${toolList.tools.map((t) => t.name).join(", ")}`
```
agentydragon: should we use our own logger or logger-wrapper that scopes everything to approval-gate? it's not very DRY this way

approval_gate/test_operator_auth.py
```diff
@@ -0,0 +1,114 @@
...
+async def test_admin_jwt_accepted_on_approve(app, admin_jwt):
+    """Valid Authentik admin JWT must be accepted (200) on approve endpoint."""
+    async with httpx.AsyncClient(transport=ASGITransport(app=app), base_url="http://test") as client:
+        response = await client.post(f"/api/actions/{_ACTION_ID}/approve", headers={"X-Authentik-Jwt": admin_jwt})
+    assert response.status_code == 200
+
+
+async def test_admin_jwt_accepted_on_reject(app, admin_jwt):
+    """Valid Authentik admin JWT must be accepted (200) on reject endpoint."""
+    async with httpx.AsyncClient(transport=ASGITransport(app=app), base_url="http://test") as client:
```
agentydragon: duplicated, extract yield fixture?

approval_gate/storage.py
```diff
@@ -0,0 +1,129 @@
...
+    updated_at: Mapped[str] = mapped_column(String, nullable=False, default=_now, onupdate=_now)
+    call_json: Mapped[str] = mapped_column(Text, nullable=False)
+    justification: Mapped[str] = mapped_column(Text, nullable=False)
+    session_key: Mapped[str | None] = mapped_column(String, nullable=True)
+    state_json: Mapped[str] = mapped_column(Text, nullable=False)
+    status: Mapped[str] = mapped_column(String, nullable=False)
+
+
+def _row_to_action(row: _ActionRow) -> Action:
+    return Action(
```
agentydragon: perhaps this should be on _ActionRow

openclaw/approval-gate/index.ts
```diff
@@ -0,0 +1,438 @@
...
+      try {
+        action = await readActionResource(client, uri);
+      } catch (err) {
+        api.logger.warn(`approval-gate: failed to read resource ${uri}: ${String(err)}`);
+        return;
+      }
+
+      // Only deliver notifications for terminal states
+      const { status } = action.state;
+      if (status !== "done" && status !== "rejected" && status !== "withdrawn") return;
```
agentydragon: refactor array contains

approval_gate/frontend/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  created_at: string;
+  updated_at: string;
+  call: ToolCall;
+  justification: string;
+  session_key: string | null;
+  state: ActionState;
+};
+
+// Wrapper types matching what api.ts returns (mirrors old REST response shapes
+// so App.svelte doesn't need changes).
```
agentydragon: No. Didn't be lazy. Propagate the refactor

approval_gate/BUILD.bazel
```diff
@@ -0,0 +1,264 @@
...
+    deps = [
+        "@pypi//fastmcp",
+        "@pypi//mcp",
+        "@pypi//pyjwt",
+        "@pypi//starlette",
+    ],
+)
+
+py_library(
+    name = "app_no_ui",
```
agentydragon: Having both app and app-no-ui which have basically the same deps feels a bit weird. I'm not even sure if we still need it? And the schema generation may also be no longer necessary?

approval_gate/conftest.py
```diff
@@ -0,0 +1,54 @@
+"""pytest configuration for approval_gate tests."""
+
+from __future__ import annotations
+
+from contextlib import asynccontextmanager
+from pathlib import Path
+
+import pytest
+
+# Patch FastMCP._docket_lifespan to a no-op for all tests in this package.
```
agentydragon: Is any of this even still needed now that we're on fastmcp 3 and using uniform fastmcp for actions?

approval_gate/mcp_auth.py
```diff
@@ -0,0 +1,104 @@
...
+
+    async def authenticate(self, conn: HTTPConnection):
+        jwt_header = conn.headers.get("x-authentik-jwt")
+        if jwt_header:
+            token_str = f"jwt:{jwt_header}"
+        else:
+            auth = conn.headers.get("authorization", "")
+            if not auth.lower().startswith("bearer "):
+                return None
+            token_str = auth[7:]
```
agentydragon: Removeprefix?

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,305 @@
...
+        public_base_url: str,
+        instructions_template_path: str,
+        **kwargs: Any,
+    ) -> None:
+        # Instructions are set after backend connection in lifespan; placeholder here
+        super().__init__("Approval Gate", instructions="Approval gate — initialising…", **kwargs)
+        # FastMCP.__init__ sets self._lifespan = default_lifespan as an instance attribute,
+        # which shadows our _lifespan class method override. Remove it so Python's method
+        # resolution finds ApprovalGateServer._lifespan instead.
+        del self.__dict__["_lifespan"]
```
agentydragon: Do we still need this hack?

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,305 @@
...
+                self._register_wrapped_tool(tool)
+
+            # ── Operator MCP tools ────────────────────────────────────────────
+            # Gated by require_scopes("operator"); bypassed for in-process clients
+            # (stdio/memory transport), which allows tests to call these freely.
+
+            @self.tool(auth=require_scopes("operator"))
+            async def list_actions(status: str | None = None, limit: int = 100) -> list[Action]:
+                """List queued/processed actions, optionally filtered by status."""
+                return await self._req_storage.list_by_status(ActionStatus(status) if status else None, limit=limit)
```
agentydragon: Just make action be the strenum already on arrival

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,305 @@
...
+            "resource-updated notifications to learn when the action completes."
+        ).strip()
+
+        async def _tool_handler(
+            justification: str,
+            input: dict[str, object] = {},  # noqa: B006
+            session_key: str | None = None,
+        ) -> ActionRef:
+            call = ToolCall(tool_name=tool_name, arguments=input)
+            action_id = str(uuid.uuid4())
```
agentydragon: If you use uuids for action IDs, make them all be consistently uuids everywhere in all tool inputs and consistently the uuid type wherever you can use it

approval_gate/test_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+    return ApprovalGateServer(
+        backend=srv,
+        db_path=tmp_path / "gate.db",
+        predicate=lambda tool, args: NeedsHumanDecision(),
+        public_base_url="http://test",
+        instructions_template_path=_INSTRUCTIONS_TEMPLATE,
+    )
+
+
+async def _wait_for_status(client: Client, action_id: str, status: ActionStatus) -> Action:
```
agentydragon: You have fastmcp notification for a reason, you should not need to write a busy poll

approval_gate/test_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+        await gate.decide(action_id, ApproveDecision())
+        async with asyncio.timeout(5.0):
+            await _wait_for_status(client, action_id, ActionStatus.done)
+    assert calls == [{"text": "hello"}]
+
+
+async def test_reject_leaves_action_rejected_and_skips_backend(gate, backend):
+    """Reject path: tool call queued → operator rejects → rejected state, backend not called."""
+    _, calls = backend
+    async with Client(gate) as client:
```
agentydragon: Looks like a dryable fixture

approval_gate/proxy_server.py
agentydragon-agent: Tried this, but FastMCP's `@self.resource` decorator validates the return type before the `ResourceContent` serialisation path — it only accepts `str`, `bytes`, or `list[ResourceContent]`, not a pydantic model. Kept `action.model_dump_json()`.

approval_gate/proxy_server.py
agentydragon-agent: Already done — the tool handler calls `self._spawn(self._apply_predicate(...))` and returns `ActionRef` immediately. `_spawn` schedules the predicate check as a background task.

approval_gate/proxy_server.py
agentydragon-agent: Same as above — the `ApproveDecision` path calls `self._spawn(self._execute_and_finish(...))` which runs the backend call in the background. The `decide()` method returns the `ExecutingState` action immediately.

approval_gate/storage.py
agentydragon-agent: Done — `ApprovalGateServer.__init__` now takes `db_path: Path` and initialises storage inside `_lifespan`. `ActionStorage.initialize` is the factory method.

approval_gate/storage.py
agentydragon-agent: Done — renamed to `list_actions(status: str | None, *, limit)` and all call sites updated directly.

approval_gate/storage.py
agentydragon-agent: Done — moved to `_ActionRow.to_action()`.

approval_gate/test_storage.py
agentydragon-agent: Removed.

openclaw/approval-gate/index.ts
agentydragon-agent: Checked the openclaw SDK — it doesn't expose a gateway connection class; plugins call `api.gatewayConnect()` and get back a raw gateway instance. Added a TODO comment in case that changes.

openclaw/approval-gate/index.ts
agentydragon-agent: Done — removed all `.slice(0, 8)` truncation; full UUIDs are used in all log/message strings now.

openclaw/approval-gate/index.ts
agentydragon-agent: Done — added `scopedLogger(api)` that returns `{ info, warn, error }` each prefixing with `approval-gate: `. `GatewayConnection` now takes a `ScopedLogger`.

openclaw/approval-gate/index.ts
agentydragon-agent: Done — extracted `TERMINAL_STATUSES = ["done", "rejected", "withdrawn"] as const` and `isTerminal()` type-guard helper.

approval_gate/test_operator_auth.py
agentydragon-agent: Done — extracted `agent_tool_names` and `operator_tool_names` as yield fixtures.

approval_gate/frontend/types.ts
agentydragon-agent: Done — removed both wrapper types. `api.ts` returns `Action[]`/`Action` directly; `App.svelte` updated to match.

approval_gate/BUILD.bazel
agentydragon-agent: Removed the `app_no_ui` target and `export_schema.py` entirely — neither is referenced externally and the REST API is gone.

approval_gate/conftest.py
agentydragon-agent: Removed — FastMCP 3's `_docket_lifespan` already no-ops when the `docket` package is absent, so the patch was redundant.

approval_gate/mcp_auth.py
agentydragon-agent: Done.

approval_gate/proxy_server.py
agentydragon-agent: Removed — pass `lifespan=self._lifespan` directly to `super().__init__()` instead, so FastMCP stores our bound method as the instance attribute from the start.

approval_gate/storage.py
agentydragon: Status should be the strenum

approval_gate/frontend/api.ts
```diff
@@ -0,0 +1,19 @@
+import { getMcpClient } from "./mcp.ts";
+import type { Action, ActionStatus } from "./types.ts";
+
+export const api = {
+  listActions: async (status?: ActionStatus, limit = 100): Promise<Action[]> => {
```
agentydragon: I think those should live on the client class itself not in a separate module

approval_gate/frontend/App.svelte
```diff
@@ -0,0 +1,70 @@
...
+  <header><h1>Approval Gate</h1></header>
+  <main><p>Loading…</p></main>
+{:else if error}
+  <header><h1>Approval Gate</h1></header>
+  <main><p class="error">Failed to load: {error}</p></main>
+{:else if actionId !== null}
+  {#if action}
+    <ActionDetail {action} />
+  {:else}
+    <header><h1>Approval Gate</h1></header>
```
agentydragon: Looks like you have some refactors/dedupes to do

approval_gate/proxy_server.py
agentydragon-agent: Done — `list_actions` now accepts `ActionStatus | None` directly; FastMCP will reject invalid status strings at the tool boundary. Updated `storage.list_actions` signature to match.

approval_gate/proxy_server.py
agentydragon-agent: Acknowledged. The current code uses `str` consistently for action IDs throughout (storage, models, tool inputs). Moving to `uuid.UUID` type would require updating `ActionRef`, `Action`, storage signatures, and tool parameter types. Worth doing but a separate refactor — keeping as typed strings for now to not mix concerns in this PR.

approval_gate/test_e2e.py
agentydragon-agent: The busy poll is covered by an `asyncio.timeout(5.0)` so it's bounded. Converting to subscription-based polling would require hooking into FastMCP's `notifications/resources/updated` handler in the Python client — the Python `Client` API does expose `set_notification_handler` but the test would need to manage an `asyncio.Event` per action ID. Happy to do this if you'd prefer it cleaner, but the current pattern is simple and correct given the 5s cap.

approval_gate/test_e2e.py
agentydragon-agent: Done — extracted `_make_gate(srv, tmp_path, predicate, db_name)` helper; both the `gate` fixture and `test_auto_approve_predicate_skips_queue` use it.

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,301 @@
...
+        self._backend_client: Client | None = None
+        self._pending_approval_lock = asyncio.Lock()
+        # Holds references to fire-and-forget background tasks to prevent GC cancellation.
+        self._background_tasks: set[asyncio.Task[Any]] = set()
+
+    # ── Lifespan: connect backend, register wrapped tools + resources ─────────
+
+    @asynccontextmanager
+    async def _docket_lifespan(self) -> AsyncGenerator[None]:
+        """No-op: the approval gate uses asyncio.create_task directly, not FastMCP's docket task system.
```
agentydragon: Do we actually need this?

openclaw/approval-gate/index.ts
```diff
@@ -0,0 +1,452 @@
...
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
+import WebSocket from "ws";
+
+const DEFAULT_GATEWAY_WS_URL = "ws://127.0.0.1:18789";
+
+// ── Action state types (mirrors approval_gate/models.py + mcp_types.CallToolResult) ─
+
+interface ActionState {
+  status: "pending" | "executing" | "done" | "rejected" | "withdrawn";
```
agentydragon, line 34: A bunch of those types are duplicated with the gate frontend. Duplicate in shared module

openclaw/approval-gate/index.ts
```diff
@@ -0,0 +1,452 @@
...
+  // The approval gate emits ResourceUpdated on resource://actions/{id} for
+  // every state change. We read the resource, format a result message, and
+  // inject it into the appropriate agent session.
+  client.setNotificationHandler(
+    { method: "notifications/resources/updated" } as Parameters<typeof client.setNotificationHandler>[0],
+    async (notification) => {
+      const uri = (notification.params as { uri?: string }).uri;
+      if (!uri?.startsWith("resource://actions/")) return;
+
+      const actionId = uri.replace("resource://actions/", "");
```
agentydragon: Deduplicate constant

openclaw/approval-gate/index.ts
agentydragon, line 1: Btw I'd like this plugin to be able to deal with the approval gate becoming available/unavailable, ideally unregistering/registering tools as they may change on reconnect

approval_gate/BUILD.bazel
```diff
@@ -0,0 +1,247 @@
...
+        "@pypi//pyjwt",
+        "@pypi//starlette",
+    ],
+)
+
+py_library(
+    name = "app",
+    srcs = ["app.py"],
+    data = [
+        "instructions.mako",
```
agentydragon: I find it surprising that two prod libraries depend on this data file and suspect it's unnecessary

approval_gate/frontend/types.ts
agentydragon-agent: Done — added `ActionId = string` type alias in `types.ts` and updated `Action.id` to use it. `mcp.ts` `approve`/`reject` methods also use `ActionId`.

approval_gate/BUILD.bazel
agentydragon-agent: Both are already gone — `export_schema.py` was deleted and `app-no-ui` / `export_schema` targets were removed from `BUILD.bazel` in the previous commit. `create_app` still has `include_static=True` as the default (used for prod) with `False` only in `test_operator_auth` to avoid needing the bundled assets during tests. I can drop that flag entirely if you'd prefer — the test fixture already doesn't call `create_app` at all (it constructs `Starlette` directly).

approval_gate/conftest.py
agentydragon-agent: Yes, still needed: the `pytest_configure` hook in `conftest.py` sets `asyncio_mode = 'auto'` and forces function-scoped event loops (`asyncio_default_fixture_loop_scope = 'function'`). Without it, async tests don't auto-detect and cross-test anyio cancel scopes leak between tests. The `storage` fixture is used by `test_storage.py`.

approval_gate/mcp_auth.py
agentydragon-agent: Already done — `mcp_auth.py` line 99 uses `.removeprefix("Bearer ")`.

approval_gate/proxy_server.py
agentydragon-agent: Yes — `FastMCP.add_tool(self, tool)` bypasses `OpenAIStrictModeMixin.add_tool`, which would reject the wrapped schema because backend schemas aren't guaranteed to have `additionalProperties: false` or all optional fields in `required`. Calling the grandparent's `add_tool` directly is the cleanest way I found to skip that validation.

approval_gate/proxy_server.py
agentydragon-agent: Done — `list_actions` in the operator tool takes `status: ActionStatus | None` (FastMCP auto-coerces the JSON string to the `StrEnum`). `approve_action`/`reject_action` take `action_id: uuid.UUID` (also auto-coerced). The resource template parameter still comes in as `str` (FastMCP doesn't coerce resource template params), so `uuid.UUID(action_id)` conversion remains there.

approval_gate/proxy_server.py
agentydragon-agent: Done — `Action.id` is now `uuid.UUID` in Python models, storage uses `str(action_id)` for SQLite, operator tools take `action_id: uuid.UUID`, `ActionRef.action_id` is `uuid.UUID`, and `ActionId = string` type alias added in the TypeScript frontend.

approval_gate/test_e2e.py
agentydragon-agent: Done — replaced busy-poll with `_ResourceWaiter(MessageHandler)`. One subtlety: calling `read_resource` from within `on_resource_updated` deadlocks (the handler runs inside `_receive_loop` which can't process its own response). The waiter just calls `event.set()` in the handler; `wait_for()` does the read from outside `_receive_loop` in a loop. Race condition is handled by registering the event before each read.

approval_gate/test_e2e.py
agentydragon-agent: Already extracted — `_make_gate(srv, tmp_path, predicate)` helper + `gate` fixture (uses `NeedsHumanDecision`). The auto-approve test creates its own gate inline since it needs a different predicate and db name, which feels appropriate.

openclaw/approval-gate/index.ts
agentydragon-agent: Addressed — there's a `scopedLogger(api)` helper that returns a `{ info, warn, error }` object prefixing all messages with `approval-gate:`. All logging goes through `const log = scopedLogger(api)`.

approval_gate/test_operator_auth.py
agentydragon-agent: Done — the duplicated connection logic is now in `agent_tool_names` and `operator_tool_names` fixtures. Tests just consume those directly.

openclaw/approval-gate/index.ts
agentydragon-agent: Done — switched from `(TERMINAL_STATUSES as readonly string[]).includes(status)` to `Set<TerminalStatus>.has()`, eliminating the awkward cast.

approval_gate/mcp_auth.py
```diff
@@ -0,0 +1,104 @@
...
+        - ``"jwt:<raw_jwt>"`` — operator Authentik JWT
+        - bare string         — agent bearer key
+        """
+        if token.startswith("jwt:"):
+            return await self._verify_operator_jwt(token[4:])
+        return self._verify_agent_bearer(token)
+
+    def _verify_agent_bearer(self, token: str) -> AccessToken | None:
+        if token == self._agent_api_key:
+            return AccessToken(token=token, client_id="agent", scopes=[])
```
agentydragon: Since agent action proposing tools should only be displayed to real agent, shouldn't we gate them behind some scope just for the sake of prevention if nothing else?

approval_gate/mcp_auth.py
```diff
@@ -0,0 +1,104 @@
...
+            )
+        except InvalidTokenError as exc:
+            logger.warning("operator JWT verification failed: %s", exc)
+            return None
+        except Exception as exc:
+            logger.error("JWT verification error: %s", exc)
+            return None
+        # Any valid JWT is treated as an operator; group checks are left to the JWKS
+        # issuer policy. If finer-grained enforcement is needed, add a groups check here.
+        return AccessToken(token=raw_jwt, client_id="operator", scopes=["operator"])
```
agentydragon: I'd prefer shared constants for those scope literals wherever they are used

approval_gate/models.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+# ── Action lifecycle states ───────────────────────────────────────────────────
+
+
+class ActionStatus(StrEnum):
+    pending = "pending"
+    executing = "executing"
+    done = "done"
+    rejected = "rejected"
```
agentydragon: Those constants should be screaming snake

approval_gate/models.py
```diff
@@ -0,0 +1,151 @@
...
+
+class WithdrawDecision(BaseModel):
+    """Agent withdrew the action before a decision was made."""
+
+    kind: Literal["withdrawn"] = "withdrawn"
+
+    model_config = ConfigDict(extra="forbid")
+
+
+OperatorDecision = Annotated[ApproveDecision | DenyDecision | WithdrawDecision, Field(discriminator="kind")]
```
agentydragon: Withdraw is not an operator decision

approval_gate/models.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+# ── Queued-action reference (returned to the agent immediately) ───────────────
+
+
+class ActionRef(BaseModel):
+    """Reference to a queued action, returned immediately when a tool call is submitted.
```
agentydragon: Just replace this by using the action id directly

approval_gate/proxy_server.py
```diff
@@ -0,0 +1,292 @@
...
+                backend_instructions=backend_instructions, public_base_url=self._public_base_url
+            )
+            # FastMCP.instructions is a property that writes through to _mcp_server.instructions
+            self.instructions = rendered_instructions
+
+            # Register a resource template for individual action state
+            @self.resource("resource://actions/{action_id}")
+            async def action_resource(action_id: str) -> str:
+                """Current state of a deferred action."""
+                uid = uuid.UUID(action_id)
```
agentydragon: I would expect taking action id as uuid directly in resource FN should work, check

openclaw/approval-gate/index.ts
agentydragon-agent: Deferred to a follow-up PR. Added a TODO comment near the MCP client setup in `openclaw/approval-gate/index.ts`:

```typescript
// TODO: handle approval gate availability changes — unregister/re-register tools on disconnect/reconnect
```

approval_gate/BUILD.bazel
agentydragon-agent: Fixed — `instructions.mako` is now loaded via a module-level `_INSTRUCTIONS_TEMPLATE = Path(__file__).parent / "instructions.mako"` in `proxy_server.py`, so only the `proxy_server` target needs it in `data`. Removed it from `app`'s `data` list in `BUILD.bazel`.

approval_gate/mcp_auth.py
agentydragon-agent: Done on both fronts:
1. `_verify_agent_bearer` now grants `AGENT_SCOPE` (`scopes=[AGENT_SCOPE]` instead of `scopes=[]`)
2. All wrapped backend tools are registered with `auth=require_scopes(AGENT_SCOPE)` on the `FunctionTool`; operator tools use `require_scopes(OPERATOR_SCOPE)`

Since FastMCP only skips auth for STDIO transport (not the memory transport used in in-process tests), I added an autouse fixture `_agent_auth_ctx` in `test_e2e.py` that injects AGENT_SCOPE into the MCP `auth_context_var` before the server task starts — anyio copies contextvars to new asyncio tasks, so the server task inherits agent scope for the duration of each test.

approval_gate/mcp_auth.py
agentydragon-agent: Done — added `AGENT_SCOPE = "agent"` and `OPERATOR_SCOPE = "operator"` as module-level constants in `mcp_auth.py`, and imported them in `proxy_server.py` for all `require_scopes()` calls.

approval_gate/models.py
agentydragon-agent: Done — `ActionStatus` members are now `PENDING`, `EXECUTING`, `DONE`, `REJECTED`, `WITHDRAWN`. String values are unchanged (`PENDING = "pending"` etc.) so JSON serialisation and DB comparisons are unaffected. Updated all usages in `Literal` type annotations, state models, and tests.

approval_gate/models.py
agentydragon-agent: Fixed — `WithdrawDecision` is removed from `OperatorDecision`; `OperatorDecision = Annotated[ApproveDecision | DenyDecision, Field(discriminator="kind")]`. The server now has a separate `withdraw()` method and a dedicated `withdraw_action` MCP tool gated by `AGENT_SCOPE`.

approval_gate/models.py
agentydragon-agent: Done — `ActionRef` class deleted; `_tool_handler` returns `uuid.UUID` directly. FastMCP serialises it as a JSON-quoted UUID string. Tests parse with `uuid.UUID(json.loads(result.content[0].text))`. The openclaw plugin similarly parses with `const actionId = JSON.parse(firstContent.text) as string`.

approval_gate/proxy_server.py
agentydragon-agent: Done — the resource function now takes `action_id: uuid.UUID` directly. FastMCP uses `validate_call` (Pydantic) on resource template functions, so the UUID string from the URI is coerced automatically — no manual `uuid.UUID()` conversion needed.

PR #824, opened 2026-03-12
Replace operator MCP tools with REST API + SSE

agentydragon:
## Summary

Replaces the MCP-based operator interface (approve_action, reject_action, list_actions tools) with a standard REST API (`/api/actions/*`) and Server-Sent Events (SSE) for live updates. The operator frontend now uses HTTP instead of MCP, simplifying the architecture and enabling better separation of concerns between agent and operator roles.

## Key Changes

- **New REST API** (`airlock/operator_api.py`): FastAPI sub-application with endpoints for listing, fetching, approving, and rejecting actions. Uses JWT validation with `decide` scope (separate from agent `propose` scope).

- **ActionCoordinator** (`airlock/coordinator.py`): Extracted shared coordination state (pending decision futures, SSE subscribers) that both MCP and REST faces operate on. Decouples the server from storage initialization.

- **Frontend migration** (`airlock/frontend/api.ts`): New REST client replacing MCP client. Implements SSE streaming with fetch + ReadableStream (supports Bearer auth headers, unlike EventSource). Provides subscription callbacks for real-time action updates.

- **DualVerifierOIDCProxy** (`airlock/oidc_auth.py`): New auth provider that validates both agent JWTs (propose scope) and operator JWTs (decide scope) against the same OIDC issuer, enabling role-based access control.

- **Test infrastructure**: Updated `conftest.py` with mock OIDC provider, `OperatorClient` HTTP client, and simplified test fixtures. Removed MCP-based operator client transport.

- **Storage schema**: Added `client_id` and `subject` columns to track action provenance (which client/user initiated the action).

## Notable Implementation Details

- **Auth separation**: Agents authenticate via MCP with `propose` scope; operators authenticate via REST with `decide` scope. Both use the same OIDC issuer and JWT validation.

- **SSE for live updates**: Frontend subscribes to `/api/events` (SSE endpoint) for real-time action state changes, eliminating the need for polling or MCP resource subscriptions.

- **Backward compatibility**: MCP endpoint remains for agents; operator tools are removed from MCP entirely (not deprecated, just moved to REST).

- **OpenAPI schema export**: New `export_openapi_schema.py` generates TypeScript type definitions for the frontend's openapi-fetch client.

- **Removed**: Claude Code OAuth2 blueprint (no longer needed with public REST API).

https://claude.ai/code/session_01UAZ2XEDXkiXZc1UGkLNW1L


airlock/oidc_auth.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Any
+
+from fastmcp.server.auth.auth import AccessToken
+from fastmcp.server.auth.oidc_proxy import OIDCProxy
+from fastmcp.server.auth.providers.jwt import JWTVerifier
+
+logger = logging.getLogger(__name__)
+
+
+class DualVerifierOIDCProxy(OIDCProxy):
```
agentydragon, line 25: this would be probably easier to write as something like a generic sequential combiner over multiple verifiers - in this case an OIDC proxy followed by JWT verifier

airlock/proxy_server.py
```diff
@@ -364,22 +347,20 @@ async def _apply_decision(
...
+                    key, ExecutingState(), ExecutionStartedDetail(started_at=started_at)
+                )
                 client = self._backend_clients.get(namespace)
                 if client is None:
                     raise RuntimeError(f"backend client not connected for namespace: {namespace!r}")
                 outcome = await client.call_tool_mcp(tool_name, input)
-                await self._update_and_notify(key, DoneState(outcome=outcome), ExecutionFinishedDetail(outcome=outcome))
+                await self.coordinator.update_and_log(
+                    key, DoneState(outcome=outcome), ExecutionFinishedDetail(outcome=outcome)
+                )
```
agentydragon: i think execution should also belong into the coordinator, rather than be owned by the mcp server.

PR #839, opened 2026-03-15
feat(kubespand): add Talos API integration layer

agentydragon-agent:
## Summary

- Add MachineService Version RPC to kubespand's gRPC server (returns kubespand version via standard Talos Version API)
- Change default API socket path to `constants.MachineSocketPath` (`/system/run/machined/machine.sock`) to match Talos machined
- Add Bazel build scaffold for Talos apid binary from `@talos_internal` source, with SELinux patch to avoid unnecessary deep deps
- Add apid wrapper binary at `cluster/kubespand/cmd/apid/`
- Add apid NixOS systemd service to kubespand module (disabled by default, `requires = kubespand.service`, opens port 50000)
- Add `apid-bin` flake input and CI release workflow updates to ship apid alongside kubespand
- Switch QEMU test `WaitForPeers` from log parsing to COSI gRPC client (queries `PeerSpec` resources directly from Unix socket)

## Architecture

Talos has two API layers: **machined** (COSI state on Unix socket) → **apid** (mTLS proxy on port 50000). The previous PR implemented the machined-equivalent layer. This PR adds:

1. **MachineService Version RPC** — `talosctl version` works against kubespand
2. **apid build scaffold** — builds upstream Talos apid from `@talos_internal` with a patch removing the SELinux dependency (not needed on non-Talos Linux)
3. **NixOS systemd unit** — `systemd.services.apid` alongside kubespand, enabled via `ducktape.kubespand.apid.enable`
4. **COSI-based test assertions** — tests now query kubespand's state directly via gRPC instead of parsing log files

Note: apid requires TLS certificate resources in COSI state to function at runtime (for mTLS). Providing those certificates is future work — this PR establishes the build and deployment infrastructure.

## Test plan

- [x] `bazel build //cluster/kubespand/...` passes (including apid binary)
- [x] `bazel test //cluster/kubespand/api:api_test` passes (Version, List, Create denied, Destroy denied)
- [ ] Full QEMU integration tests with COSI-based WaitForPeers
- [ ] NixOS module evaluation (requires nix flake check)

https://claude.ai/code/session_01G1kRYfMYzhAAL1ddjr8do8


cluster/kubespand/qemu_tests/vms/kubespanlib/kubespanlib.go
```diff
@@ -53,6 +61,8 @@ func ConfigureNetwork(linkIP, linkMask string) {
 func StartKubespand(cfg KubespandConfig) *exec.Cmd {
 	os.MkdirAll("/var/lib/kubespan", 0o755)
 	os.MkdirAll("/etc/kubespan", 0o755)
+	// Create socket directory for kubespand's COSI API server.
+	os.MkdirAll("/system/run/machined", 0o700)
```
agentydragon, line 65: shouldn't kubespand do this if it's missing? same with /var/lib/kubespan?

cluster/kubespand/kubespand.example.yaml
```diff
@@ -99,8 +99,9 @@ cluster:
 #   port: 7445
 
 # api:
-#   # Unix socket path for the read-only COSI state gRPC server.
+#   # Unix socket path for the COSI state + MachineService gRPC server.
```
agentydragon, line 102: we don't have to have most of this config commented, it's just an example.

nix/nixos/modules/kubespand.nix
```diff
@@ -40,6 +46,28 @@ in
       default = false;
       description = "Enable debug logging";
     };
+
+    apid = {
+      enable = lib.mkEnableOption "apid (Talos API proxy on port 50000 with mTLS)";
+
+      package = lib.mkOption {
```
agentydragon, line 53: can we make them both be in the same package? easier

PR #840, opened 2026-03-15
PostToolUse hook: check-only lint with diff feedback

agentydragon-agent:
## Summary

- Change PostToolUse hook from auto-formatting files to **check-only**: runs pre-commit, captures output and diff, then **restores** the original file so the agent can fix issues itself
- Report lint failures back to the agent via `additional_context` with truncated pre-commit output and a short unified diff of what pre-commit would change
- Rename `SESSION_START` constant to `HOOK_DISPATCH` in test helpers (it was already pointing at hook_dispatch)
- Add `data = [":bazel_wrapper"]` to hook_dispatch binary target

## Test plan

- [x] `bazel build //devinfra/claude:post_tool_use_lib` — builds with mypy
- [x] `bazel test //devinfra/claude:test_post_tool_use` — 18 tests pass (guard tests, git root, serialization, diff generation, format output, pre-commit integration with mocks for: diff included, no file change, passes, timeout restores, not found, multiple hooks fail)

https://claude.ai/code/session_014QTw213uphnU7tY7APzxBm

### DIFF
--- a/devinfra/claude/BUILD.bazel
+++ b/devinfra/claude/BUILD.bazel
@@ -341,6 +341,7 @@ py_library(
 
 py_binary(
     name = "hook_dispatch",
+    data = [":bazel_wrapper"],
     imports = ["../.."],
     main_module = "devinfra.claude.hook_dispatch",
     visibility = ["//visibility:public"],
--- a/devinfra/claude/post_tool_use.py
+++ b/devinfra/claude/post_tool_use.py
@@ -1,13 +1,24 @@
-"""PostToolUse hook: auto-format files after Edit/Write tool calls.
+"""PostToolUse hook: lint-check files after Edit/Write tool calls.
 
-Runs pre-commit on files that Claude Code just modified. Delegates all
-formatter dispatch to pre-commit's own configuration (.pre-commit-config.yaml),
-maintaining a single source of truth for file-to-formatter mapping.
+Runs pre-commit on files that Claude Code just modified, reports any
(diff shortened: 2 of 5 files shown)


devinfra/claude_hooks/hook_dispatch.py
```diff
@@ -1,43 +1,57 @@
...
 
 def main() -> None:
     raw = sys.stdin.read()
 
     try:
-        event = json.loads(raw).get("hook_event_name", "")
-    except json.JSONDecodeError:
+        hook_input = _hook_adapter.validate_json(raw)
+    except Exception:
         return
```
agentydragon: Do not swallow parse errors. Let them crash.

devinfra/claude_hooks/testing/shell_helpers.py
```diff
@@ -7,7 +7,7 @@
 from pathlib import Path
 
 # Runfiles paths for binaries
-SESSION_START = "_main/devinfra/claude_hooks/session_start"
+SESSION_START = "_main/devinfra/claude_hooks/hook_dispatch"
```
agentydragon: This is no longer session start only

agentydragon-agent:
Fixed both comments in 4b92fe7:
- Removed the try/except around `validate_json` — parse errors now propagate to the outer error handler
- Renamed `SESSION_START` → `HOOK_DISPATCH` in `shell_helpers.py` and updated the reference in `session_start_helpers.py`

PR #845, opened 2026-03-15
Refactor cluster validation from monolithic script to modular test suite

agentydragon:
## Summary

Refactored the cluster validation system from a single monolithic `validate_cluster` binary into a modular test suite with individual Bazel test targets. This enables better test isolation, clearer failure reporting, and integration with standard testing infrastructure.

## Key Changes

- **Removed monolithic binary**: Deleted `main.py` and the `validate_cluster` binary that ran all checks sequentially
- **Created modular test targets**: Split validation into 7 independent test targets under `test_live_*`:
  - `test_live_orphaned_files` — YAML files must be referenced by kustomizations
  - `test_live_blueprint_completeness` — Authentik blueprint files listed in configMapGenerator
  - `test_live_dependencies` — No circular dependencies, required dependencies present
  - `test_live_controller_health_checks` — HelmReleases/Terraform CRs have healthChecks
  - `test_live_kustomize` — No duplicate external-secrets, no CRD layering violations
  - `test_live_flux_build` — Flux build succeeds and produces expected resources
  - `test_live_helm_templates` — Cilium Helm values render without errors
- **Introduced kustomize build serialization**: New `kustomize_build_all` binary and `kustomize_build_results` genrule that:
  - Builds all kustomizations once in parallel
  - Serializes results to JSON for consumption by downstream tests
  - Avoids redundant kustomize invocations across multiple test targets
- **Added shared test fixtures** (`live_conftest.py`): Provides `cluster`, `k8s_dir`, `workspace`, and `kustomize_build_results` fixtures for live tests
- **Created `cluster/k8s:all_yaml` filegroup**: Exposes all YAML files as a data dependency for pure-analysis tests
- **Moved checks to test modules**: Extracted `find_orphaned_files` and `check_duplicate_external_secrets` from `checks.py` into their respective test files
- **Updated precommit integration**: Removed cluster validation from precommit hooks (now runs via Bazel test suite)
- **Added unit test for serialization** (`test_check_runner.py`): Validates round-trip JSON serialization of build results

## Implementation Details

- All live tests tagged with `cluster_validation` for easy filtering: `bazel test --test_tag_filters=cluster_validation //cluster/...`
- Pure-analysis tests use `//cluster/k8s:all_yaml` as data deps (sandboxed)
- Kustomize-dependent tests consume pre-built results from genrule to avoid redundant builds
- Tests use pytest fixtures for dependency injection and session-scoped caching
- Kustomize results serialized with relative paths for portability across environments

https://claude.ai/code/session_013Tt625tndDneBPxP9CXXSU


cluster/k8s/authentik/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+# gazelle:ignore
+
+load("//devinfra/testing:defs.bzl", "py_test")
+
+py_test(
+    name = "test_blueprint_completeness",
+    srcs = ["test_blueprint_completeness.py"],
+    data = ["//cluster/k8s:all_yaml"],
```
agentydragon: this just needs the yamls under *here*, no?

cluster/scripts/validate_cluster/BUILD.bazel
```diff
@@ -1,15 +1,16 @@
 # gazelle:ignore
 # Cluster validation scripts
 #
-# Two binaries:
-#   validate_cluster — kustomizations + helm templates (unified, fast local checks)
+# Binaries:
```
agentydragon: no need for this overarching comment, you can just put that with each directive/section if useful

cluster/validation/dependencies.py
```diff
@@ -19,7 +19,7 @@ class _DependencyRule:
 
 _DEPENDENCY_RULES: list[_DependencyRule] = [
     # external-secrets-config ordering is enforced dynamically by validate_external_secrets_dependencies().
-    # vault → external-secrets ordering is enforced by CRD layering checks in checks.py.
+    # vault → external-secrets ordering is enforced by CRD layering checks in test_kustomize.py.
```
agentydragon, line 22: should this file and related checks also go into a test_ ?

cluster/scripts/validate_cluster/kustomize.py
```diff
@@ -59,9 +59,10 @@ def parse_kustomize_file(kust_file: Path) -> KustomizeFile | None:
         return KustomizeFile(path=kust_file, resources=resources, patches=patches)
 
 
-async def run_kustomize_build(kustomization_path: Path) -> KustomizeBuildResult:
+async def run_kustomize_build(kustomization_path: Path, *, kustomize_bin: Path | None = None) -> KustomizeBuildResult:
```
agentydragon: i don't think we need to allow None? what's the None path even used for? also isn't this only called in the genrule? should we inline it?

cluster/scripts/validate_cluster/test_flux_build.py
```diff
@@ -0,0 +1,28 @@
...
+
+pytest_plugins = ["cluster.scripts.validate_cluster.conftest"]
+
+
+def test_flux_build_succeeds(k8s_dir: Path) -> None:
+    result = run_flux_build(k8s_dir)
+    assert result.returncode == 0, f"flux build failed:\nstdout:\n{result.stdout}\nstderr:\n{result.stderr}"
+    assert result.stdout.strip(), f"flux build returned empty output:\nstderr: {result.stderr.strip() or 'none'}"
+
+    resources = list(parse_k8s_resources(yaml.safe_load_all(result.stdout)))
```
agentydragon: inline resources + no need for list call.

cluster/terraform/bootstrap/infrastructure/test_helm_templates.py
```diff
@@ -0,0 +1,21 @@
+"""Test: Cilium Helm values files render without errors."""
```
agentydragon, line 1: move this test to live in bootstrap/infrastructure

cluster/scripts/validate_cluster/test_kustomize.py
```diff
@@ -0,0 +1,264 @@
...
+    "Vault": "vault-operator",
+    # tofu-controller
+    "Terraform": "tofu-controller",
+    # powerdns-operator
+    "ClusterZone": "powerdns-operator",
+    "ClusterRRset": "powerdns-operator",
+}
+
+# These Kustomizations ARE the operators, so they don't need to depend on themselves
+OPERATOR_KUSTOMIZATIONS = {
```
agentydragon: can't you derive that from the above?

cluster/scripts/validate_cluster/test_kustomize.py
```diff
@@ -0,0 +1,264 @@
...
+        assert (
+            _check_crd_layering(
+                _make_build_result(Path("/k8s/test-app/overlays/production/kustomization.yaml"), output)
+            )
+            == []
+        )
+
+
+class TestCrdToOperatorMapping:
+    def test_common_crds_mapped(self) -> None:
```
agentydragon: delete this test, this is useless

cluster/k8s/test_kustomize.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_no_duplicate_external_secrets() -> None:
+    """Exactly one external-secrets HelmRelease must exist."""
+    deployments: dict[tuple[str | None, str | None], list[Path]] = {}
+    for result in _RESULTS:
+        for resource in result.resources:
+            if resource.kind == "HelmRelease" and resource.name == "external-secrets":
+                key = (resource.namespace, resource.chart_version)
+                deployments.setdefault(key, []).append(result.kustomization_path.parent)
```
agentydragon: defaultdict?
also, would this make sense to generalize to any HelmRelease resource name?

cluster/validation/BUILD.bazel
```diff
@@ -127,40 +97,75 @@ py_binary(
...
-        "@multitool//tools/tofu",
-    ],
-    imports = ["../../.."],
-    main_module = "cluster.scripts.validate_cluster.sealed_secrets",
-    deps = [":sealed_secrets"],
-)
-
 py_library(
     name = "helm_templates",
     srcs = ["helm_templates.py"],
```
agentydragon, line 98: testonly?

cluster/validation/crd_layering.py
```diff
@@ -0,0 +1,73 @@
...
+
+# Derived: CRD kind -> operator name (for error messages)
+CRD_TO_OPERATOR: dict[str, str] = {
+    kind: operator for operator, kinds in OPERATOR_CRDS.items() for kind in kinds
+}
+
+
+def check_crd_layering(result: KustomizeBuildResult) -> list[str]:
+    """Check if a kustomization mixes HelmReleases with CRD instances."""
+    if not result.success:
```
agentydragon: KustomizeBuildResult should not have a 'success' bool, it should only represent successful builds. if kustomize build comes out of a genrule, failure means the genrule will fail and the test won't even run.

cluster/validation/crd_layering.py
```diff
@@ -0,0 +1,73 @@
...
+    has_helmrelease = any(r.kind == "HelmRelease" for r in result.resources)
+    crd_instances = [(r.kind, CRD_TO_OPERATOR[r.kind]) for r in result.resources if r.kind in CRD_TO_OPERATOR]
+
+    if has_helmrelease and crd_instances:
+        kust_name = result.kustomization_path.parent.name
+        unique_crds = sorted({f"{k} (needs {op})" for k, op in crd_instances})
+        return [
+            f"{kust_name}: mixes HelmRelease with CRD instances: {', '.join(unique_crds)}. "
+            f"Split into a separate '{kust_name}-secrets/' Kustomization."
+        ]
```
agentydragon: make this function raise exception instead of returning list of readable strings

cluster/validation/crd_layering.py
```diff
@@ -0,0 +1,73 @@
...
+
+    return []
+
+
+def make_build_result(path: Path, yaml_output: str) -> KustomizeBuildResult:
```
agentydragon: a bit strange to see this here. isn't this a test-only method for the test of crd layering validation?

cluster/validation/flux.py
```diff
@@ -0,0 +1,64 @@
...
+class FluxKustomization(BaseModel):
+    """Parsed flux-kustomization.yaml Kustomization CR."""
+
+    model_config = ConfigDict(populate_by_name=True, extra="ignore")
+
+    name: str
+    file_path: Path
+    spec_path: str = Field(default="", alias="path")
+    depends_on: list[DependsOn] = Field(default=[], alias="dependsOn")
+    health_checks: list[HealthCheck] = Field(default=[], alias="healthChecks")
```
agentydragon: use snake case pydantic alias generation feature

cluster/validation/flux.py
```diff
@@ -0,0 +1,64 @@
...
+            if not doc.get("apiVersion", "").startswith("kustomize.toolkit.fluxcd.io"):
+                continue
+
+            metadata = doc.get("metadata", {}) or {}
+            name = metadata.get("name", "")
+            if not name:
+                continue
+
+            spec = doc.get("spec", {}) or {}
+            results.append(FluxKustomization.model_validate({"name": name, "file_path": flux_file, **spec}))
```
agentydragon: split FluxKustomization into {name, file, structured spec pydantic} so the spec pydantic matches the actual structure of the actual yaml

cluster/validation/flux.py
```diff
@@ -0,0 +1,64 @@
...
+    with flux_file.open() as f:
+        for doc in yaml.safe_load_all(f):
+            if not doc:
+                continue
+            if doc.get("kind") != "Kustomization":
+                continue
+            if not doc.get("apiVersion", "").startswith("kustomize.toolkit.fluxcd.io"):
+                continue
+
+            metadata = doc.get("metadata", {}) or {}
```
agentydragon: inline metadata

cluster/validation/helm_templates.py
```diff
@@ -45,12 +45,12 @@ def ensure_cilium_repo() -> bool:
...
     )
     if result.returncode != 0:
-        print(f"❌ Failed to add Cilium Helm repo: {result.stderr}")
+        print(f"Failed to add Cilium Helm repo: {result.stderr}")
         return False
 
     result = subprocess.run([_helm_bin(), "repo", "update"], check=False, capture_output=True, text=True)
     if result.returncode != 0:
-        print(f"⚠️  Failed to update Helm repos: {result.stderr}")
+        print(f"Failed to update Helm repos: {result.stderr}")
```
agentydragon: just run the above with check=True, don't swallow errors. ditto with Cilium above. you can signal errors with exceptions

cluster/validation/kustomize_build_all.py
```diff
@@ -0,0 +1,80 @@
...
+        kustomize_bin,
+        "build",
+        kustomization_path.parent,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+    stdout, stderr = await proc.communicate()
+
+    if proc.returncode != 0:
+        return KustomizeBuildResult(kustomization_path=kustomization_path, success=False, error=stderr.decode())
```
agentydragon: switch this subprocess to run sync not async, and check=True instead of returning an error object

cluster/validation/sealed_secrets.py
```diff
@@ -70,19 +66,15 @@ def main() -> int:
...
-        print(f"⚠️  No tofu state found at {tf_dir}/terraform.tfstate")
-        print("   Skipping SealedSecret validation (state not initialized)")
+        print(f"No tofu state at {tf_dir}/terraform.tfstate — skipping SealedSecret validation")
         return 0
 
-    private_key = get_private_key_from_tofu(tf_dir)
-    if not private_key:
-        print("⚠️  Could not read private key from tofu state")
-        print(f"   Run 'tofu apply' in {tf_dir} first")
+    if not (private_key := get_private_key_from_tofu(tf_dir)):
```
agentydragon: i'd expect the above to compute the state path and this to take the state path rather than the dir?

cluster/k8s/test_kustomize.py
agentydragon-agent: Done — switched to `defaultdict(list)` and generalized to check all HelmRelease names for duplicates (not just external-secrets).

cluster/validation/BUILD.bazel
agentydragon-agent: Done in previous commit — `dependencies`, `health_checks`, and `crd_layering` are now `testonly = True`.

cluster/validation/crd_layering.py
agentydragon-agent: Done — removed `success` and `error` fields from `KustomizeBuildResult`. It now only represents successful builds. `kustomize_build_all` uses `check=True` and fails the genrule on errors.

cluster/validation/crd_layering.py
agentydragon-agent: Done — `check_crd_layering` now raises `CrdLayeringViolation` instead of returning a list of strings.

cluster/validation/crd_layering.py
agentydragon-agent: Moved to `test_crd_layering.py` as a module-level `_build_result()` helper.

cluster/validation/flux.py
agentydragon-agent: Done — using `pydantic.alias_generators.to_camel` on `FluxKustomizationSpec`.

cluster/validation/flux.py
agentydragon-agent: Done — split into `FluxKustomizationSpec` (matches YAML spec shape with camelCase alias generation) and `FluxKustomization` (name + file_path + spec). Properties on FluxKustomization delegate to spec for backward compatibility.

cluster/validation/flux.py
agentydragon-agent: Done — inlined metadata access with walrus operator: `if not (name := (doc.get("metadata") or {}).get("name")):`

cluster/validation/helm_templates.py
agentydragon-agent: Done in previous commit — `validate_helm_template` now runs with `check=True` and raises `CalledProcessError` on failure.

cluster/validation/kustomize_build_all.py
agentydragon-agent: Done — switched to sync `subprocess.run` with `check=True`. Removed asyncio entirely.

cluster/validation/sealed_secrets.py
agentydragon-agent: Done — `get_private_key_from_tofu` now takes `state_path: Path` directly. Uses `-state` flag for tofu output and `check=True`.

cluster/validation/flux.py
```diff
@@ -0,0 +1,78 @@
...
+    @property
+    def spec_path(self) -> str:
+        return self.spec.path
+
+    @property
+    def depends_on(self) -> list[DependsOn]:
+        return self.spec.depends_on
+
+    @property
+    def health_checks(self) -> list[HealthCheck]:
```
agentydragon: let's remove those @properties, have callers use x.spec.whatever

cluster/validation/test_crd_layering.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _build_result(path: Path, yaml_output: str) -> KustomizeBuildResult:
+    return KustomizeBuildResult(
+        kustomization_path=path, resources=parse_k8s_resources(yaml.safe_load_all(yaml_output))
+    )
+
+
+class TestCrdLayeringCheck:
+    def test_valid_helmrelease_only(self) -> None:
```
agentydragon: parameterized tests?

cluster/k8s/authentik/BUILD.bazel
agentydragon-agent: Yes, `glob(["**/*.yaml"])` in `cluster/k8s/authentik/BUILD.bazel` already only captures yamls under the `authentik/` directory — Bazel globs are relative to the package. This filegroup is used by the blueprint completeness test, not by kustomize builds (those use `all_kustomize_inputs` which includes non-YAML files).

cluster/scripts/validate_cluster/BUILD.bazel
agentydragon-agent: Done — removed all section banner comments from `cluster/validation/BUILD.bazel`.

cluster/validation/dependencies.py
agentydragon-agent: It's already `testonly = True` in BUILD.bazel. It's a validation library with its own unit tests (`test_dependencies.py`), used by `test_cluster_validation.py`. Keeping it as a separate module allows the unit tests to exercise the graph/cycle/rule logic independently of the full cluster fixture. Moving the functions into a test file would lose that separation.

cluster/scripts/validate_cluster/kustomize.py
agentydragon-agent: This file was in the old `cluster/scripts/validate_cluster/` location which has been deleted. The kustomize build logic is now in `cluster/validation/kustomize_build_all.py` — it takes no None path, uses `check=True`, and is only invoked by the genrule.

cluster/scripts/validate_cluster/test_flux_build.py
agentydragon-agent: Done — the current version at `cluster/k8s/test_flux_build.py` inlines the set comprehension directly: `kinds = {r.kind for r in parse_k8s_resources(yaml.safe_load_all(flux_build_output))}`.

cluster/terraform/bootstrap/infrastructure/test_helm_templates.py
agentydragon-agent: Done — it already lives at `cluster/terraform/bootstrap/infrastructure/test_helm_templates.py`.

cluster/scripts/validate_cluster/test_kustomize.py
agentydragon-agent: This was on the old `cluster/scripts/validate_cluster/test_kustomize.py` which has been deleted. The current version at `cluster/k8s/test_kustomize.py` loads results once at module level and parametrizes the CRD layering check over them.

cluster/scripts/validate_cluster/test_kustomize.py
agentydragon-agent: Done — deleted. Build failures are caught by the genrule itself (exits non-zero).

cluster/validation/flux.py
agentydragon-agent: Done — removed all three `@property` delegations. Callers now use `kust.spec.depends_on`, `kust.spec.health_checks`, `kust.spec.path` directly. Updated `dependencies.py`, `health_checks.py`, `test_flux.py`, `test_dependencies.py`, and `test_health_checks.py`.

cluster/validation/test_crd_layering.py
agentydragon-agent: Done — consolidated all valid-case tests into a single `@pytest.mark.parametrize` with descriptive IDs (`helmrelease_only`, `crd_only`, `operator_with_crd`, `cert_manager_nested`, `vault_nested`, `overlay_skipped`). The violation test remains separate since it uses `pytest.raises`.

cluster/validation/dependencies.py
agentydragon, line 108: i think this could generalize to all the known operators and CRDs, not just external secrets. and we should also allow it if it depends just transitively, via the dependency graph building we have.

agentydragon, line 141: this is suspicious. if graph failed to build, we should just raise error, not just continue.

agentydragon, line 147: use exceptions to signal errors

cluster/validation/kustomize_build_all.py
```diff
@@ -0,0 +1,60 @@
...
+
+    results = [_run_kustomize_build(k) for k in kustomization_files]
+
+    # Relativize kustomization_path to k8s_dir for portability
+    for r in results:
+        r.kustomization_path = r.kustomization_path.relative_to(k8s_dir)
+
+    adapter = TypeAdapter(list[KustomizeBuildResult])
+    output_path.write_text(adapter.dump_json(results, indent=2).decode())
+    return 0
```
agentydragon, line 56: don't return int

cluster/validation/test_crd_layering.py
```diff
@@ -0,0 +1,120 @@
...
+    apiVersion: helm.toolkit.fluxcd.io/v2
+    kind: HelmRelease
+    metadata:
+      name: vault-webhook
+    ---
+    apiVersion: vault.banzaicloud.com/v1alpha1
+    kind: Vault
+    metadata:
+      name: vault
+""")
```
agentydragon, line 95: let's make a testdata/ and move the yaml content there from those inline multiline blobs

PR #846, opened 2026-03-15
feat(kubespand): migrate talos dep from http_archive+overlay to go_deps.module

agentydragon-agent:
## Summary

- Replaces `http_archive` + `BUILD.overlay.bazel` for `github.com/siderolabs/talos` with `go_deps.module` + `go_deps.gazelle_override` in `third_party/go.MODULE.bazel`
- `bazel build //cluster/kubespand/...` now succeeds
- Removes `third_party/siderolabs/talos_internal/` (old http_archive approach)

## Key technical choices

- `gazelle:go_visibility //visibility:public` — overrides Gazelle's hardcoded restricted visibility for Go `internal/` packages, making talos internals accessible from kubespand
- `gazelle:exclude` directives for files that require problematic dependencies (vip/etcd, ethernet API mismatch due to go.mod replace directive, dns upstream/resolve files)
- `gazelle:proto disable_global` for `containerd/cgroups/v3` and `siderolabs/siderolink` — includes pre-generated `.pb.go` files instead of trying to compile proto
- Cascade k8s version upgrades required by talos v1.13: apimachinery/api/client-go v0.32→v0.35, cbor v2.7→v2.9, kube-openapi, gnostic-models v0.6→v0.7

## Test plan

- [x] `bazel build //cluster/kubespand/...` succeeds
- [x] `//cluster/kubespand/api:api_test` passes
- [x] `//cluster/kubespand/qemu_tests/nft:nft_test` passes
- [x] `//cluster/kubespand/qemu_tests/talos:talos_test` passes
- [ ] `//cluster/kubespand/qemu_tests/kubespan:kubespan_test` — VM networking timeout (environment issue, not code regression)
- [ ] `//cluster/kubespand/qemu_tests/doublenat:doublenat_test` — VM networking timeout (environment issue, not code regression)

https://claude.ai/code/session_013dpuryfsDpXoMGruzPPbr3


cluster/kubespand/kubespand.example.yaml
```diff
@@ -99,8 +99,22 @@ cluster:
...
-#   # Unix socket path for the read-only COSI state gRPC server.
+#   # Unix socket path for the COSI state + MachineService gRPC server.
 #   # Exposes all kubespand resources (peer status, affiliates, identity, etc.)
-#   # for querying via COSI gRPC clients.
-#   # Default: /var/run/kubespand/kubespand.sock
-#   socket_path: "/var/run/kubespand/kubespand.sock"
+#   # for querying via COSI gRPC clients, plus Version RPC for talosctl.
+#   # Uses the same path as Talos machined so apid can proxy it.
+#   # Default: /system/run/machined/machine.sock
+#   socket_path: "/system/run/machined/machine.sock"
```
agentydragon, line 107: Didn't we remove the override option

cluster/kubespand/qemu_tests/kubespan/kubespan_test.go
```diff
@@ -24,6 +26,105 @@ func TestDiscoveryOnly(t *testing.T) {
...
+	// Wait for Talos CP API (boot takes ~60-120s on TCG).
+	talosClient := h.NewTalosClient(t, h.RunfilePath(t, h.TalosConfig), fmt.Sprintf("127.0.0.1:%d", talosAPIPort))
+	defer talosClient.Close()
+	h.WaitForTalosAPI(t, talosClient, "192.168.50.2", 180*time.Second)
+	sw.Lap("Talos CP API ready")
+
+	// Connect to kubespand's apid — success proves the full chain:
+	// kubespand → OSRootController → APICertSANsController → APIController
+	// → trustd CSR → secrets.API → apid serves mTLS on :50000.
+	kubespandClient := h.NewTalosClient(t, h.RunfilePath(t, h.TalosConfig), fmt.Sprintf("127.0.0.1:%d", kubespandAPIPort))
```
agentydragon, line 102: If we have a concept of a kubespand/talos VM, we could give them the management port and maybe Talos config too and make this more dry

cluster/kubespand/qemu_tests/talos_helpers.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// PollKubeSpanStatus polls the Talos COSI API for KubeSpan peer status.
+func PollKubeSpanStatus(t *testing.T, c *client.Client, nodeIP string, timeout time.Duration) ([]KubespanPeerResult, error) {
+	t.Helper()
+
+	deadline := time.Now().Add(timeout)
+	var lastErr string
+
+	for time.Now().Before(deadline) {
```
agentydragon: I suspect with this encapsulated "wait for x" functions you may be understanding go channel capabilities and I'm noticing their shape is duplicated

cluster/kubespand/qemu_tests/vms/trustd/main.go
```diff
@@ -0,0 +1,78 @@
...
+			Message: "missing required kernel cmdline params",
+			Error:   fmt.Sprintf("cluster_id=%s discovery=%s ca_crt_len=%d token=%s endpoint=%s", clusterID, discoveryAddr, len(caCrtB64), token, clusterEndpoint),
+		})
+		initlib.Poweroff()
+	}
+
+	caCrtPEM, err := base64.StdEncoding.DecodeString(caCrtB64)
+	if err != nil {
+		initlib.EmitEvent(qemu_tests.Event{Type: qemu_tests.EventError, Message: "ca_crt base64 decode failed", Error: err.Error()})
+		initlib.Poweroff()
```
agentydragon: Should we just defer this?

cluster/kubespand/qemu_tests/vms/trustd/main.go
```diff
@@ -0,0 +1,78 @@
...
+	cfg := kubespanlib.KubespandConfig{
+		ClusterID:       clusterID,
+		SharedSecret:    sharedSecret,
+		DiscoveryAddr:   discoveryAddr,
+		ListenPort:      51820,
+		EndpointFilters: []string{"192.168.50.0/24"},
+		ClusterEndpoint: clusterEndpoint,
+		CACrt:           string(caCrtPEM),
+		Token:           token,
+	}
```
agentydragon, line 66: Can't we instead bake kubespand config into initramfs or a partition? I'm basically thinking of making the init side thin and reducing boilerplate we add for tests

cluster/kubespand/kubespand.example.yaml
agentydragon-agent: Still there — only the default changed (from `/var/run/kubespand/kubespand.sock` to `/system/run/machined/machine.sock` so apid can use it without reconfiguration). The override option is still valid if you need a different path.

cluster/kubespand/qemu_tests/kubespan/kubespan_test.go
agentydragon-agent: Good call. Added `MgmtNIC(port int) []string` parallel to `McastNIC` — both return `[]string` of QEMU args. `BootTalosVM` now uses `MgmtNIC` internally, and the test constructs the kubespand VM with `append(h.McastNIC(...), h.MgmtNIC(kubespandAPIPort)...)` instead of the inline netdev args.

The kernel args still live in the test since `talosConfig` is test-package-local. A further step could be a `BootKubespandTrustdVM(t, clusterID, sharedSecret, ..., apiPort)` helper once there's a second test needing the same setup.

cluster/kubespand/qemu_tests/talos_helpers.go
agentydragon-agent: Agreed on duplication. Added `pollUntil(deadline time.Time, fn func() bool) bool` which handles the loop+sleep, and refactored both `WaitForTalosAPI` and `PollKubeSpanStatus` to use it. The channel-based `WaitForEvent`/`RequireEvent` pattern already handles event-driven waits; these two are polling against an external RPC so the deadline loop is appropriate, just needed to not repeat it.

cluster/kubespand/qemu_tests/vms/trustd/main.go
agentydragon-agent: Restructured with `run() error` — all error cases `return fmt.Errorf(...)`, and `main()` calls poweroff in one place. `defer` at the top of `main()` doesn't work because the success path blocks on `select{}` forever, so the defer would never fire. The `run()` extraction achieves the same goal: poweroff is written once.

cluster/kubespand/qemu_tests/vms/trustd/main.go
agentydragon-agent: Good direction long-term. It would make the init binary a thin shim (`LoadConfig("/kubespand.yaml")` instead of 6 cmdline args) and eliminate the base64-encode-the-PEM hack. The blocker is test-side: we'd need to dynamically append a file to the CPIO archive at test time (or use a second virtio data drive like the Talos CIDATA pattern). Happy to do that in a follow-up once this lands.

PR #851, opened 2026-03-16
Support multiple build targets per release package

agentydragon:
## Summary
Refactored the release configuration system to support multiple build targets per package, enabling packages like `kubespand` to release multiple binaries (e.g., `kubespand` and `apid`) in a single release workflow.

## Key Changes

- **Configuration Model**: Introduced `ReleaseTarget` class to represent individual build targets with optional Nix flake inputs. Changed `ReleaseConfig.bazel_target` (single string) to `ReleaseConfig.targets` (list of `ReleaseTarget` objects).

- **Release Workflow Generation**: Updated `generate_consolidated_release()` to:
  - Build and prepare multiple binaries/wheels in a single release job
  - Generate dynamic step names based on artifact count ("binary" vs "binaries")
  - Collect all distribution paths and create release files from multiple targets
  - Handle flake input updates for each target independently

- **Release Decision Logic**: Modified `check_release.py` to check if *any* of a package's targets have been affected, rather than checking a single target. The check-release-needed action now accepts space-separated `bazel_targets` instead of a single `bazel_target`.

- **Bazel Label Parsing**: Enhanced `BazelLabel.parse()` to support short-form labels (e.g., `//foo/bar` → `//foo/bar:bar`), eliminating the need for explicit target names when they match the package name.

- **Configuration Updates**: Migrated all release packages in `workflows.yaml` to the new `targets` structure, with `kubespand` now releasing both the main daemon and the `apid` proxy daemon.

## Implementation Details

- Flake input updates are now per-target, allowing different targets to update different Nix inputs
- The release body for `kubespand` was updated to document both binaries
- All generated workflow steps properly handle multiple targets with appropriate pluralization
- Backward compatibility maintained through the short-form label parsing enhancement

https://claude.ai/code/session_01QKfNyE79m3wCHyjSoZCRfP

### DIFF
--- a/devinfra/ci/generate_ci.py
+++ b/devinfra/ci/generate_ci.py
@@ -191,14 +191,12 @@ def _pkg_id(name: str) -> str:
     return name.replace("-", "_")
 
 
-def _binary_dist_path(bazel_target: str) -> str:
+def _binary_dist_path(label: BazelLabel) -> str:
     """Compute the bazel-bin path for a binary target.
 
     rules_go appends a _ suffix to the output directory for go_binary targets.
     """
-    label = BazelLabel.parse(bazel_target)
-    target_name = label.name
-    return f"bazel-bin/{label.package}/{target_name}_/{target_name}"
+    return f"bazel-bin/{label.package}/{label.name}_/{label.name}"
 
 
 def generate_consolidated_release(releases: dict[str, ReleaseConfig]) -> Workflow:
@@ -224,7 +222,7 @@ def generate_consolidated_release(releases: dict[str, ReleaseConfig]) -> Workflo
                 uses="./.github/actions/check-release-needed",
                 with_args={
                     "package_prefix": name,
-                    "bazel_target": config.bazel_target,
+                    "bazel_targets": " ".join(str(t.bazel_target) for t in config.targets),
                     "latest_release_tag": f"{name}-latest",
                 },
             )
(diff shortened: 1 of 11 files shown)


devinfra/ci/generate_ci.py
```diff
@@ -354,20 +354,24 @@ def generate_consolidated_release(releases: dict[str, ReleaseConfig]) -> Workflo
             )
 
         if config.artifact_type == "binary":
-            dist_path = _binary_dist_path(config.bazel_target)
-            label = BazelLabel.parse(config.bazel_target)
+            target_strs = [t.bazel_target for t in config.targets]
+            labels = [BazelLabel.parse(t) for t in target_strs]
```
agentydragon: could we make bazel_target already be parsed as BazelLabel, if pydantic has that option?

util/bazel/query.py
```diff
@@ -67,7 +63,12 @@ def parse(cls, s: str) -> BazelLabel:
             raise ValueError(f"Invalid Bazel label (must start with {_PKG_SEP!r} or {_REPO_SIGIL!r}): {s!r}")
 
         if _TARGET_SEP not in rest:
-            raise ValueError(f"Invalid Bazel label (no {_TARGET_SEP!r} separator): {s!r}")
+            # Support short form: //foo/bar implies //foo/bar:bar
+            package = rest
+            name = Path(package).name
+            if not name:
+                raise ValueError(f"Invalid Bazel label (no {_TARGET_SEP!r} separator and empty package): {s!r}")
+            return cls(repo=repo, package=Path(package), name=name)
```
agentydragon: just make package be Path?

PR #940, opened 2026-03-20
Fix e2e test: remove proxy env_inherit that breaks RBE

agentydragon-agent:
## Summary
- Remove proxy-related `env_inherit` vars (`HTTPS_PROXY`, `SSL_CERT_FILE`, etc.) from the container e2e test BUILD.bazel
- These leaked host proxy env vars to RBE workers where the referenced CA file paths don't exist, causing `FileNotFoundError` in `_build_upstream_proxy_args`
- The test container sets up its own proxy infrastructure and doesn't need host proxy config
- Keep only `DOCKER_HOST` which is needed for Docker daemon access

Also includes prior commits from PR #939:
- Selective log collection via `docker exec` instead of bind-mounts (fixes root-ownership permission issues)
- Bump test size to `large` and add Python-level timeout for graceful cleanup

## Test plan
- [x] `bazel test //devinfra/claude/testing/container_e2e:test_container_e2e --spawn_strategy=remote` passes (verified twice, ~98-100s)

https://claude.ai/code/session_01LKkSj4NqVScSgSyduphUgL

### DIFF
--- a/devinfra/claude/testing/BUILD.bazel
+++ b/devinfra/claude/testing/BUILD.bazel
@@ -15,7 +15,6 @@ py_library(
         "//devinfra/claude/auth_proxy:setup",
         "//devinfra/claude/auth_proxy:vars",
         "@pypi//cryptography",
-        "@pypi//pydantic",
     ],
 )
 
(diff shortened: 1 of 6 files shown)


agentydragon-agent:
Superseded by #941 which was merged with these changes.

PR #1324, opened 2026-04-16
reverse_engineer skill: build-what's-missing doctrine + known-defeateated

agentydragon:
- Add opening principle: when a tool fails, diagnose the exact check that
  failed, write the minimal fix, add it to examples/, record it in the skill
- Add "Defeating Obfuscation" subsection with the 5-step diagnostic pattern
  and a known-defeated techniques table seeded with the three garble issues
  we've already solved: pclntab magic XOR (pclntool), .symtab stripping
  (use GNU objdump -d), and module info stripping (go version -m → unknown)
- Frame pclntool as an instance of the pattern (GoReSym also confirmed broken
  on garble magic XOR) rather than a one-off tool

BAZEL_TEST_INVOCATIONS=none: documentation-only change
https://claude.ai/code/session_01TRvwMDg9D6Yz6n8N8bFNTh


skills/reverse_engineer/examples/garble_re_recipe.sh
```diff
@@ -74,23 +87,16 @@ fns=()
...
 done
 
 # strings[0] and [1] are from connectToServer → same garbled function
 [ "${fns[0]}" = "${fns[1]}" ] \
-  || {
-    echo "FAIL: expected same function for connectToServer strings (got ${fns[0]} vs ${fns[1]})"
-    exit 1
-  }
+  || fail "expected same function for connectToServer strings (got ${fns[0]} vs ${fns[1]})"
 echo "PASS: connectToServer strings map to the same garbled function (${fns[0]})"
```
agentydragon: no need to have paired separate pass/fail messages, having the fail only printed with one final all-pass is fine. this example is intended to be read by an agent anyway so it'll see which fails are not being hit.

skills/reverse_engineer/examples/pclntool.go
```diff
@@ -30,49 +37,125 @@ var goMagics = []uint32{
...
+	if err != nil {
+		die(".gopclntab read: %v", err)
+	}
+	fileOffset := sec.Offset
+	textAddr := f.Section(".text").Addr
+
+	origMagic := binary.LittleEndian.Uint32(pclntabData[:4])
+	magic, _, ok := findWorkingMagic(pclntabData, textAddr)
+	if !ok {
+		die("could not find working pclntab magic")
```
agentydragon: panic?

agentydragon: panic?

skills/reverse_engineer/examples/pclntool.go
```diff
@@ -30,49 +37,125 @@ var goMagics = []uint32{
 	0xfffffffb, // Go 1.2
 }
 
+func die(format string, a ...any) {
```
agentydragon: maybe we can just replace this withpanic

PR #1345, opened 2026-04-18
grocy_mcp: tool surface cleanup + case-driven eval harness + Haiku rollouts

agentydragon:
## Summary

Two logical chunks on top of the previous Haiku-eval work:

**1. Tool surface cleanup (`289cd11`)** — doc + schema pass across every
custom batch tool, triggered by friction the first Haiku rollout surfaced.
- Split the entity-type enum into `WriteableEntityType` (used by
  `create_entities`) and `ReadableEntityType` (superset, used by
  `list_entities` + `get_entities`) so `_view` / `_resolved` / computed
  entities can be read but not created. Also drops the `shopping_list`
  items entity in favour of the typed shopping-list tools.
- Replaces singular `create_product` with batch `create_products`; adds
  typed batch `create_locations` + `create_quantity_units` so reference-
  data bootstrap uses the same shape as every other batch tool.
- Makes `EntityResolver` fully stateless (every resolve re-fetches) and
  hoists one shared instance in `register_batch_tools` to drop
  per-tool boilerplate.
- Second-person imperative docstrings across every batch tool with
  consistent conventions (`Name or ID.`, `YYYY-MM-DD`, the QU-must-match
  constraint wherever `qu` appears) and cross-links between siblings.
- Expanded `EditProductField` / `EditStockEntryField` / `EditShoppingListField`
  with per-enum rationale for which fields are nullable.
- `test_e2e` gets a bootstrap test over the new batch creators, a
  read-vs-write entity-type test, and a docstring cross-link smoke test
  that fails if any back-tick reference in a tool doc doesn't resolve
  to a registered tool.

**2. Case-driven eval harness + rollouts (`cd704e2`, `fdd40c5`, `e7cf458`)** —
broadens the eval from a single hardcoded prompt to a registry of cases
with initial-state seeds and prose success criteria.
- `cases.py` registers three scenarios: `populate_fresh_pantry` (the
  existing bootstrap, now wrapped), `shopping_planning` ("I'm going
  shopping, planning to cook X, what do I need on top of the existing
  list?"), `post_cook_logging` ("I just cooked X, used all of Y,
  N left of Z").
- `seed.py` populates each case's initial Grocy state over the REST API
  — realistic lived-in pantry with ~30 products (bolognese-relevant
  ingredients at mixed stock levels plus ~18 unrelated background items
  as negative examples).
- `run.py` snapshots final Grocy state (`/stock`, shopping list, stock
  log + reference tables for name resolution) to `final_state.json`
  alongside the transcript and summary.
- `TODO.md` captures the rest of the scenarios we want eventually
  (consume + location disambiguation, expiry cleanup, shopping-list
  round-trip, pantry audit, QU-conversion stress, etc.) with one-line
  descriptions of what tools each exercises.

**Rollouts (`e7cf458`)** — all three cases against
`claude-haiku-4-5-20251001`, artifacts under
`x/grocy_mcp/eval/sample_rollouts/`. Real findings surfaced:
- **shopping_planning**: Haiku guessed `"default"` as the shopping-list
  name, got a silent failure, and never added anything to the list —
  points at `get_shopping_list` needing a default argument or a much
  better error message.
- **post_cook_logging**: Haiku consumed most ingredients correctly but
  undercounted the onion (`consume_stock` with delta=1 when delta=2
  would match "one onion left"). Argues for preferring `inventory_stock`
  for "N left" framings in the tool docs.
- **populate_fresh_pantry**: postmortem flags date-format strictness on
  `add_stock` and the Fridge-already-exists error when pre-seeded data
  is present.

**Supporting fixes:**
- `GROCY_MCP_HOST_NETWORK=1` toggle on the Grocy container for gvisor
  sandboxes where Docker port publishing is a no-op.
- `wait_for_grocy_ready` now actually waits for Grocy's lazy SQLite
  migrations to finish: hit `/` first to trigger them, then poll
  `/api/objects/locations` until it returns a JSON array.
- Disable four OpenAPI-generated tools (`list_users`,
  `list_product_locations`, `list_product_stock_entries`,
  `list_location_stock`) whose `query[]` filter parameter violates
  Anthropic's tool-schema property-key regex. TODO to re-expose as
  hand-written batch tools.
- Fix an empty-transcript bug: `HistoryProvider`s store messages under
  `session.state[source_id]`, not `session.state` directly.

## Test plan

- [x] `bbr test //x/grocy_mcp:test_e2e //x/grocy_mcp:test_server //x/grocy_mcp:test_config //x/grocy_mcp:test_fix_openapi_spec //x/grocy_mcp:test_resolver //x/grocy_mcp:test_retry_safety` all green (invocations `339be85b-eb07-4508-81e5-02ef85800a77` / `bbab36f0-8f15-437b-a08b-61862c9554ce`).
- [x] `bb run //x/grocy_mcp:dump_tools` prose-check across every custom batch tool.
- [x] Live Haiku rollout of all three cases; artifacts checked in under `sample_rollouts/`.


x/grocy_mcp/eval/run.py
```diff
@@ -79,37 +79,109 @@ def _write_messages_jsonl(messages: list[Message], path: Path) -> None:
...
+    # Store paths relative to the summary so the artifacts are portable
+    # (checked into the repo, read from any working dir, etc).
+    result = EvalResult(
+        case_id=case.id,
+        success_criteria=case.success_criteria,
+        model=model,
+        api=api,
+        postmortem_text=postmortem_text,
+        transcript_path=transcript_path.relative_to(output_dir),
+        final_state_path=final_state_path.relative_to(output_dir),
```
agentydragon: Don't store those paths, they are obvious

PR #1350, opened 2026-04-19
ci(release-artifact): fail fast on bazel_output / artifacts.py filename mismatch

agentydragon:
## Summary

Tonight's `claude-hook-rs` release uploaded an asset named `claude_hook` (Bazel `rust_binary` crate naming), but `artifacts.py` registered `claude-hook`. sync-pins caught this 30 minutes later with a 404 from the download URL. Move the check to release time so the failure is immediate and obvious.

Parses `artifacts.py` via regex (no new entry point). Fails the release-artifact action step if `basename(bazel_output) != ARTIFACTS[pkg].filename`, with an actionable message.

## Test plan

- [x] Next `claude-hook-rs` release build: guard no-ops (filename matches)
- [x] If someone changes the Bazel target output name without updating `artifacts.py`, CI fails fast instead of producing an unsyncable pin

https://claude.ai/code/session_013kEZA6hzAzB7s2rnnHibbk

### DIFF
--- a/.github/actions/release-artifact/action.yml
+++ b/.github/actions/release-artifact/action.yml
@@ -54,6 +54,18 @@ runs:
           exit 1
         fi
 
+        # Verify basename(bazel_output) matches devinfra/ci/artifacts.py
+        # filename for this pkg. sync-pins fetches the release asset by
+        # that exact filename; a mismatch (e.g. rust_binary "claude_hook"
+        # vs registered "claude-hook") yields 404 at next sync. Catch
+        # here at release time, not 30 min later at sync-pins time.
+        expected_filename=$(uv run --script devinfra/ci/artifacts.py filename "$PKG")
+        actual_filename=$(basename "$artifact")
+        if [[ "$expected_filename" != "$actual_filename" ]]; then
+          echo "::error::Artifact filename mismatch for $PKG: Bazel produced '$actual_filename', artifacts.py registers '$expected_filename'. Fix one or the other so sync-pins can find the release asset."
+          exit 1
+        fi
+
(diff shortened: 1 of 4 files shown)


.github/actions/release-artifact/action.yml
```diff
@@ -54,6 +54,23 @@ runs:
...
 
+        # Verify basename(bazel_output) matches devinfra/ci/artifacts.py
+        # filename for this pkg. sync-pins fetches the release asset by
+        # that exact filename; a mismatch (e.g. rust_binary "claude_hook"
+        # vs registered "claude-hook") yields 404 at next sync. Catch
+        # here at release time, not 30 min later at sync-pins time.
+        expected_filename=$(python3 -c "
+        import re, sys
+        src = open('devinfra/ci/artifacts.py').read()
+        m = re.search(r'pkg=\"' + sys.argv[1] + r'\", filename=\"([^\"]+)\"', src)
```
agentydragon: i'm not a fan of parsing python by regex

agentydragon: Fair. Replaced with a proper CLI on `artifacts.py`:

```bash
expected_filename=$(uv run --script devinfra/ci/artifacts.py filename "$PKG")
```

`artifacts.py` now has PEP 723 script metadata + a `filename <pkg>` subcommand that does a real typed lookup via `ARTIFACTS`. Moved `sources_path()` (the only thing that needed `util.bazel.workspace`) into `sync_pins.py` so `artifacts.py` stays deployable as a standalone `uv run --script` with just `pydantic` + `httpx`.

---

PR #1462, opened 2026-05-03
props/frontend/debundle: live-browser smoke target (blocked on svelte-data-table chunk-fold)

agentydragon:
## Summary

Realistic-shape live-browser smoke for `props/frontend` (Svelte 5 / esbuild splitting / Tailwind), self-hosted, exercising the same debundle pipeline shape as the Tana smoke but on an open-source corpus.

**Status:** branch published, smoke does NOT pass yet — blocked on a real (preexisting) debundler bug uncovered while authoring the spec. Filing this PR so the wiring + spec + test infra get reviewed and the blocker has a public reference.

## What's wired up under `props/frontend/debundle/`

- `BUILD.bazel` — `:smoke_bundle`, `:snapshot`, `:debundle_props_frontend` (using the now-shared `debundle_pipeline`), `:load_props_frontend` (`js_test`).
- `esbuild.smoke.config.mjs` — multi-chunk smoke variant of the production esbuild config (forces vendor packages into separate shared chunks via secondary entry points).
- `prepare_snapshot.mjs` — extracts `js-files.txt`, `asset-summary.json`, `vendor-chunks.json` from the smoke metafile; renames vendor chunks to deterministic filenames so target/exec build configs agree.
- `spec_generator.mjs` — emits the canonical pipeline (apply_vendor_annotations + rename_vendor_exports + swap_vendor_chunks + materialize_logical_modules + define_logical_module + rewrite_chunk_entry_specifiers + emit_browser_harness).
- `smoke_main.ts` / `smoke_shell.svelte` / `smoke_index.html` — entry + UI shell mirroring production `App.svelte`.
- `vendor_marker.ts` / `vendor_highlight_marker.ts` / `vendor_datatable_marker.ts` — secondary entries that force esbuild's chunk-graph algorithm to put each vendor package in its own shared chunk.
- `static_load_test.mjs` — self-hosted Node http server + headless Chromium driver.

## Smoke assertions (literal)

1. `failedRequests.length > 0` → fail (no failed asset requests).
2. `consoleErrors.length > 0 || pageErrors.length > 0` → fail (no console / page errors).
3. `pageState.navText && pageState.headerText.includes("Props")` (`<nav>` + `<header h1>` carry expected text).
4. `await page.click('nav a[href*="/runs"]'); await page.waitForFunction(() => location.hash === "#/runs", ...);` plus DOM `data-debundle-smoke="pathname"` text equals `/runs` — clicking a nav link updates URL hash + DOM (no-backend interaction).

## Vendor packages picked

- `highlight.js` (`named-from-module-default` wrapper shape — `import x from "..."; export default x;`)
- `@careswitch/svelte-data-table` (no wrapper — plain re-export passthrough)

## Materialized modules

- `lib/router` (parseHash, createRouter, pathname, searchParams, goto, resolve, parseParams)
- `lib/token` (needsToken, getToken, setToken, clearToken, captureTokenFromUrl, onAuthFailed)
- `lib/highlighting` (highlightLines)

## Renames (5)

`resolve` → `resolveRoute`, `goto` → `gotoRoute`, `parseParams` → `parseRouteParams`, `parseHash` → `parseRouteHash`, `highlightLines` → `highlightCodeLines`.

## Blocker — `swap_vendor_chunks` rejects svelte-data-table

```
swapVendorChunks operation mark_vendor_datatable export shape mismatch for @careswitch/svelte-data-table@0.6.3:
vendor exports not found upstream=[append,child,derived,each,enable_legacy_mode_flag,first_child,from_html,
get,html,index,init,mount,next,pop,push,reset,set_attribute,set_class,set_text,setup_stores,sibling,
store_get,template_effect,writable]
```

Esbuild's chunk-graph algorithm folds Svelte runtime helpers into the same shared chunk as `@careswitch/svelte-data-table`. The check correctly notices the chunk re-exports more names than upstream — but leaves the user without a next step. Filing this as a separate bug with a failing-fixture e2e (work in a follow-up PR per the "minimal e2e first" discipline).

## TODO.md update

Replaces the Excalidraw entry with a forward-looking "props/frontend live-browser smoke" section describing the actual built target + workflow rule + open follow-ups.

## Test plan

- [x] `bazelisk build //props/frontend/debundle:debundle_props_frontend --remote_executor=""` — pipeline runs through `apply_vendor_annotations` + `rename_vendor_exports` + first vendor's `swap_vendor_chunks`.
- [ ] `bazelisk test //props/frontend/debundle:load_props_frontend --remote_executor=""` — fails on chunk-fold; tracked separately.
- [ ] After chunk-fold fix, smoke turns green.

https://claude.ai/code/session_01HG7aincxRPqXwaZH532PJG

---
_Generated by [Claude Code](https://claude.ai/code/session_01HG7aincxRPqXwaZH532PJG)_

### DIFF
--- a/devinfra/js/debundle/TODO.md
+++ b/devinfra/js/debundle/TODO.md
@@ -3,112 +3,66 @@
 Forward-looking gaps in the Rust debundler. Items are written to be removed
 once closed; this file is not a changelog.
 
-## Excalidraw live-browser smoke
+## props/frontend live-browser smoke
 
-Build an open-source analog of gaffer's `tana/re/web/live_proxy:load_*`
-test inside ducktape, against a Bazel-managed Excalidraw bundle. The
-motivation: when a Tana debundler issue surfaces (proxy crash, AST
-corruption, missing chunk, emit shape regression, optimisation
-behaviour bug), reproducing the failure on Excalidraw lets us share
-the repro in a public bug report, write a regression test that runs
-in ducktape's open CI, and avoid leaking proprietary Tana bundle
-detail. Excalidraw is open-source and broadly representative of "real
-React + vendored chunks + dynamic imports + service worker."
+The open-source analog of gaffer's `tana/re/web/live_proxy:load_*`
+lives at `//props/frontend/debundle:load_props_frontend`. It drives
+the canonical pipeline (`apply_vendor_annotations` →
+`rename_vendor_exports` → `swap_vendor_chunks` →
+`materialize_logical_modules` → identifier renames →
+`rewrite_chunk_entry_specifiers` → `emit_browser_harness`) against a
+multi-chunk smoke bundle of `props/frontend`'s real Svelte source,
+serves the harness with a tiny self-hosted Node http server (no
+MITM — `props/frontend` has no server-side dependency), and drives
+headless Chromium through it.
 
-### Bundle build
-
-Use a Bazel-managed Excalidraw build. Two viable paths:
-
-- **Pull a prebuilt deploy** (snapshot a specific `excalidraw.com`
-  publish into `tana/x/upstream/excalidraw/snapshots/<version>/`
-  — analogous to `tana/upstream/web/snapshots/`). Requires keeping
-  the snapshot fresh enough that upstream's auxiliary endpoints (if
-  any) still work. Easier to bootstrap.
-- **Build from source under Bazel** — Excalidraw's `excalidraw-app/`
-  builds with Vite; reproduce that build (npm + vite via
-  aspect_rules_js, or via a `genrule` shelling to npm) and feed the
-  output into the pipeline. More work upfront, but the bundle is
-  reproducible from a single git pin and we control the optimisation
-  level / minifier settings.
-
-Either way, the build configuration must roughly match Tana's: minify
-on, scrambled identifiers, production-tree-shaken, real chunk-split
-boundaries. A development build (with un-mangled names and source
-maps inlined) won't exercise the debundler's RE-relevant code paths.
-
-### Spec scope
-
-Not the round-trip minimum — the spec should exercise realistic-ish
-module extraction and rename paths, the same shape the Tana spec
-runs. Concretely:
-
-- `apply_vendor_annotations` + `swap_vendor_chunks` over a couple of
-  Excalidraw's actual vendor chunks (React, Roughjs, Pointers, etc.)
-  so vendor-swap edge cases (named-from-default,
-  named-from-module-default, default-only, JSON-default) get covered
-  on a real bundle, not only synthetic fixtures.
-- `materialize_logical_modules` over a handful of pre-identified
-  Excalidraw source modules — pick ones whose shape is recognisable
-  in the compiled output (a clearly-bounded React component, a pure
-  geometry helper, a state slice). Goal: prove the materialiser
-  recovers approximately the right symbols/files from a real
-  scrambled bundle.
-- A small set of `define_logical_module` rename operations on
-  identifiers visible in the compiled output. Goal: exercise the
-  rename pipeline at the same aggressiveness Tana uses.
-- `rewrite_chunk_entry_specifiers` + `emit_browser_harness` so the
-  output is a runnable app the live proxy can serve.
-
-The exact module list / rename list is part of the implementation —
-pick stable shapes that are unlikely to drift wildly when Excalidraw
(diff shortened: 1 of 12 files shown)


agentydragon:
This feels overcomplicated. If we cleanly reused the built frontend bundle, I would expect that to take less code.

agentydragon:
Agreed. The complexity comes from the agent forcing each vendor package into its own shared chunk via secondary entry markers (`vendor_marker.ts`, `vendor_highlight_marker.ts`, `vendor_datatable_marker.ts`) plus a separate `esbuild.smoke.config.mjs` that's a near-clone of the production config. The smoke also ships its own UI shell (`smoke_main.ts` / `smoke_shell.svelte`) instead of reusing `App.svelte`.

The simpler shape is to treat `//props/frontend:bundle` as the snapshot input directly: derive `asset-summary.json` from the metafile esbuild already emits, vendor-swap whatever happens to land in shared chunks (or nothing if the production config doesn't split out anything cleanly — that's a useful constraint on the spec, not a problem), and reuse `index.html` + `App.svelte`. If the production config doesn't yield isolatable vendor chunks, that's a finding worth reflecting in the spec rather than working around with synthetic markers.

Putting this PR on hold and dispatching a simplification agent to rebuild the smoke against the existing bundle. Will close + supersede or push a clean simplification on top.

---

agentydragon:
Also it looks like this props frontend bundle build does not actually do the same sort of transformations that the app we're building debundler for does. For example the spec you've made seems like the bundle already contains readable names.

agentydragon:
Right — the agent's renames (`resolve` → `resolveRoute`, `goto` → `gotoRoute`, `parseParams` → `parseRouteParams`, etc.) are renaming already-readable names to more descriptive names; they're not exercising the scrambled-letter-pair → readable-name path that the Tana RE workflow actually runs. So the smoke isn't a realistic-shape verification of that pipeline path even before the chunk-fold blocker.

Three options to make the corpus match:

1. **Make props/frontend's debundle pipeline run against a minified variant of its own bundle.** Esbuild's `minify: true` does mangle local identifiers, but exported top-level names appear to be preserved — would need a stricter pass (terser, swc-minify, or esbuild + `mangleProps: true` with care). Adds a secondary build config but keeps `props/frontend` as the corpus.
2. **Pick a different open-source corpus that already matches Tana's shape** — a small Vite-built React app whose production output has letter-pair scrambled identifiers, vendored chunks, dynamic imports, etc. Candidates: a small subset of Excalidraw, a Svelte-based Vite app, or a fixture we maintain.
3. **Author a synthetic corpus under `devinfra/js/debundle/e2e/` that runs Vite/terser internally** to produce a bundle with the right shape. More controlled but loses the "real world" angle; might be the right level for most coverage with the props/frontend smoke filling in the actual-real-app role only after (1) lands.

I'd lean toward (1) on the existing target — keeps the corpus open-source, lets the same `props/frontend` build act as both production and debundle test, doesn't add a new external pin. Want me to take that path, or prefer (2) or some hybrid?

---

agentydragon:
Acknowledging the user's two earlier comments + Copilot's structural review. The current branch has both fundamental issues:

1. **Overcomplicated** (separate esbuild config, three vendor markers, smoke-specific UI shell duplicating App.svelte).
2. **Bundle doesn't match Tana's shape** — names aren't scrambled, so the renames in the spec rename already-readable identifiers to other readable identifiers, not exercising the scrambled-letter-pair → readable path.
3. **Spec shape mismatches the pipeline contract** (Copilot caught: `define_logical_module` uses wrong field shape; rename ops have no `operation` field; ordering of `swap_vendor_chunks` vs `rewrite_chunk_entry_specifiers` likely wrong; `pruneOtherChunks` default drops vendor chunks; `requires-network` tag wrong on a loopback test).

Closing this PR. The simplification follow-up dispatches a fresh agent against:

- A separate **`props/frontend/debundle:tana_standin_bundle`** built specifically as a Tana-shape standin — minified with mangled top-level identifiers (terser, swc-minify, or aggressive esbuild settings), real chunk-split, vendored. The production `//props/frontend:bundle` stays as-is for the actual app.
- Reuse the production `App.svelte` + `index.html` as the smoke source. No separate UI shell.
- Spec authored with the real `define_logical_module` / vendor / rename schema (not invented shapes).
- The smoke pipeline drives the standin bundle and asserts the rename pipeline actually exercises scrambled→readable.

Will close + supersede with a new PR.

---

PR #1552, opened 2026-05-14
debundle: drop ResidualEntry variant; residual becomes a synthesized logical module

agentydragon:
## Summary

- Collapses `ModuleId` from a two-variant enum (`Logical(LogicalModuleIndex) | ResidualEntry`) to a single-field struct wrapping `LogicalModuleIndex`. Every `Partition` caller now threads a `default_destination: ModuleId`; no more implicit `ResidualEntry` sentinel anywhere in the analyzer.
- The residual catch-all is now an ordinary `LogicalModule` flagged `residual: true`. The materializer synthesizes a schedule-only `<chunk>::anon_residual_sentinel` logical module past `module_plans.len()` for use as `Partition::new`'s `default_destination`, preserving the pre-refactor split where named-unclaimed bindings and anonymous fallback statements live in distinct dep-graph destinations.
- Supersedes the `claude/catchall-file-anon-stmts` band-aid branch (just pushed, no PR yet). Both attempt to fix the same gaffer-spec anon-stmt-with-cross-module-dep symptom but this branch tackles the underlying variant-as-special-case debt.

## Why stop at one commit

The plan called for three commits:

1. **Variant removal** (landed, `4a3f111b6`): `ModuleId` collapse, `Partition::new` API change, sentinel preservation. `bbr test //devinfra/js/debundle/...` is green (35/35).
2. **Always synthesize residual + anon-stmt routing** (deferred): widen `logical_requests_for_chunk` to synthesize unconditionally for `InlineInEntry` + `CatchallFile`; sweep unclaimed anonymous side-effect statements into the residual module's `anonymous_statement_ordinals`.
3. **Peelability cleanup** (deferred): consolidate the supernode-tied call sites the Stage 3 subagent flagged.

I prototyped commit 2 (~210 LoC) but the anon-stmt sweep produces module-quotient cycles that didn't exist when anonymous statements stayed in entry. The cycle shape: an anonymous statement reads a binding owned by an explicit logical module → after the sweep, residual depends on that module at-init; the explicit module also reads a residual-owned binding at-init → cycle. In the pre-refactor world the anon stmt lived in entry, which isn't a tracked module in the dep graph. Tests under `unassigned_mode: catchall_file` (the suite default) hit this in ~11 cases. The `chunk_renames`-applies-in-entry tests also regressed because bindings the lowerer used to rewrite in place now live in the synthesized residual plan instead.

Per the task's stop-signal rule ("10 tests still failing after commit 2 and I can't figure out why quickly → stop and report") I reverted commit 2 rather than ship regressions. Commit 1 alone is the variant-removal half and is independently valuable — every analysis-layer call site now uses the field-access syntax and there's a single uniform residual destination.

## Suspicions for the commit-2 follow-up

- The validator's `has_init_order_constraining_edge` cycle test may need a relaxation for residual modules whose only cross-module edges come from anonymous side-effect statements (those statements are guaranteed to run after the named bindings they read, via the residual module's import order on entry).
- Or: anon stmts should only sweep into residual when they have no cross-module init reads, leaving the rest in entry's body. That preserves OLD CatchallFile semantics for the failure-prone case while still claiming the easy ones (the gaffer-spec shape).
- For `chunk_renames`, the lowerer needs to apply renames to bindings whose decl moves into the synthesized residual plan, not just those staying in entry's body. Today the `cross_module_chunk_renames` filter (`binding_assignment.contains_key`) excludes them.
- The new e2e test for `unassigned_mode: catchall_file` + anon-stmt-Sequenced-bound-to-named-binding (the gaffer bug shape) is also deferred; it should drive the design of whichever relaxation lands.

## Tests still failing on the band-aid PR (`claude/catchall-file-anon-stmts`)

For reference, that branch failed: `anonymous_statement_test`, `catchall_anon_statement_test`, `chunk_rename_cross_module_test`, `comma_list_owner_split_test`, `cross_module_emission_test`, `lowering_test`, `naturalization_test`, `owner_graph_purity_reason_test`, `purity_test`, `realizability_test`, `residual_member_rename_test`, `source_order_emit_test`. This PR's commit-2 prototype had a different but overlapping failure set.

## Test plan

- [x] `bbr test //devinfra/js/debundle/...` — 35/35 green on commit 1 (invocation `f478fe12-3d05-4f7a-a869-05d5bbd8a602`).

https://claude.ai/code/session_01B2sF75vhtHFb3t9y2LuZEk

---
_Generated by [Claude Code](https://claude.ai/code/session_01B2sF75vhtHFb3t9y2LuZEk)_

### DIFF
--- a/devinfra/js/debundle/analysis_tests.rs
+++ b/devinfra/js/debundle/analysis_tests.rs
@@ -193,13 +193,23 @@ Ro([Z], C.prototype, dynamicKey, 2);"#,
     }
 
     fn logical(idx: usize) -> ModuleId {
-        ModuleId::Logical(LogicalModuleIndex(idx))
+        ModuleId(LogicalModuleIndex(idx))
+    }
+
+    /// Test-helper residual destination: the synthesized residual
+    /// logical module always sits at the highest index appended by
+    /// `schedule_for` / `schedule_with_residual_module`. Tests use this
+    /// instead of the historical `ModuleId::ResidualEntry` literal.
+    fn residual() -> ModuleId {
+        ModuleId::logical(usize::MAX)
     }
 
     fn render(id: ModuleId) -> String {
-        match id {
-            ModuleId::Logical(LogicalModuleIndex(idx)) => format!("mod_{idx}"),
-            ModuleId::ResidualEntry => "<residual>".to_string(),
+        let LogicalModuleIndex(idx) = id.0;
+        if idx == usize::MAX {
+            "<residual>".to_string()
+        } else {
+            format!("mod_{idx}")
         }
     }
 
@@ -234,7 +244,8 @@ Ro([Z], C.prototype, dynamicKey, 2);"#,
         binding_assignment.insert("A".to_string(), logical(0));
         binding_assignment.insert("B".to_string(), logical(1));
         let owner_graph = build_owner_graph(&facts);
-        let partition = Partition::from_binding_assignment(&owner_graph, &binding_assignment);
+        let partition =
+            Partition::from_binding_assignment(&owner_graph, &binding_assignment, residual());
         let graph = build_module_quotient(&owner_graph, &partition);
         let report = validate_schedule(&graph, &render);
         assert_eq!(report.cycles.len(), 1);
@@ -250,7 +261,8 @@ Ro([Z], C.prototype, dynamicKey, 2);"#,
         binding_assignment.insert("B".to_string(), logical(1));
         binding_assignment.insert("C".to_string(), logical(2));
         let owner_graph = build_owner_graph(&facts);
-        let partition = Partition::from_binding_assignment(&owner_graph, &binding_assignment);
+        let partition =
+            Partition::from_binding_assignment(&owner_graph, &binding_assignment, residual());
         let graph = build_module_quotient(&owner_graph, &partition);
         let report = validate_schedule(&graph, &render);
         assert!(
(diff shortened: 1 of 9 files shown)


agentydragon:
Rebased onto current `devel` (`357ba03ed`). New head: `a319730ff`.

## Conflicts resolved

Three files had conflicts; the other six (`factor_assembly.rs`, `ids.rs`, `logical_modules.rs`, `partition.rs`, `peelability.rs`, `reports.rs`, `schedule.rs`) auto-merged.

- **`devinfra/js/debundle/peel_factorize.rs`**: devel was rewritten by #1545 and #1556 (greedy agglomeration + diagnostic dedup). The PR's edits to this file were obsolete — devel's current `peel_factorize` already discriminates the residual module via `node.destination.id == RESIDUAL_ENTRY_MODULE_ID` (string-based), which is compatible with the new struct-shape `ModuleId(LogicalModuleIndex)`. Took devel's version wholesale.
- **`devinfra/js/debundle/analysis_tests.rs`** `schedule_for` helper: devel renamed `analyze_chunk_facts(&module, &BTreeSet::new())` → `analyze_facts(&module)`. Kept devel's call and the PR's struct-shape destructuring (`let LogicalModuleIndex(i) = id.0;`). Dropped the now-redundant early `bindings.insert` from devel since the post-conflict block re-fills `bindings` after computing `residual_idx`.
- **`devinfra/js/debundle/report_schema.rs`**: devel added a new `FactorizeDiagnosticReason` enum and `FactorizeDiagnostic` struct just above `RESIDUAL_ENTRY_MODULE_ID`; the PR rewrote the constant's doc comment to reflect that residual is no longer a discriminator. Kept both: devel's new types stay, PR's updated doc comment lands on the constant.

Also verified `factorize.rs`, `validation.rs`, `graph.rs`, `lib.rs`, `spec.rs`, `e2e/chunk_renames_test.rs` use `ModuleId` only as a type (no enum-variant constructors or pattern matches), so they need no changes under the new struct shape.

## Tests

`SKIP=prettier bbr test //devinfra/js/debundle:all` — all 14 test targets pass on RBE (invocation `e36cfe52-2557-48c8-9d4e-a3c94f0374f6`).

## Scope

Stayed at commit 1 (variant removal) only, as the PR body documents. Commit 2 ("always synthesize residual + anon-stmt routing") remains deferred.

---

PR #1700, opened 2026-05-27
debundle/peel: incremental topological order (Pearce-Kelly) for cycle check

agentydragon:
## Summary

Replaces the cone-DFS in `peel::quotient::QuotientGraph::merge_creates_new_constraining_cycle` with a Pearce-Kelly 2007 incremental topological order maintained alongside the class adjacency. The cycle check reduces to a PK-style affected-region forward DFS, bounded by `|Δ|` instead of `|cone|`.

### What's new

- **`devinfra/js/debundle/peel/topo_order.rs`** — new private module with `TopoOrder { ord: Vec<u32>, is_dag: bool }`. Public surface: `init_from_dag` (Kahn), `would_create_cycle` (PK reachability), `apply_contract` (post-merge local Kahn restart over the affected window).
- **`peel::quotient::QuotientGraph`** — carries a `topo_ord` field. Initialized in both `from_report` constructors after `rebuild_class_adjacency`. Maintained incrementally on every `merge_classes_unchecked`.
- **Fallback path** — when the initial class graph contains cycles (`is_dag = false`), the kernel falls back to the original cone-DFS, preserving pre-PK behavior bit-for-bit on cyclic inputs.

### Algorithm notes

A new multi-class cycle through the merged class exists iff there is a directed path of length ≥ 2 between `c1` and `c2` in the pre-merge DAG. A direct `c1 → c2` edge just becomes a self-loop after merge and is **not** a new cycle — matching the original code's behavior (which dropped self-loops in its frontier).

In a DAG, only one direction is possible (the one from lower `ord` to higher `ord`). The forward DFS from `lo`'s out-neighbors (excluding `hi` to skip the direct-edge case), bounded by `ord ≤ ord[hi]`, gives the PK affected-region search. On the test corpus the average visited count was **0** — the affected region is empty for almost every check.

`apply_contract` is a local Kahn restart over the window `[ord[winner], ord[loser]]` — bounded by the window size, never global.

### Why I rolled my own instead of adopting `incremental-topo`

I audited the `incremental-topo` crate (declanvk/incremental-topo, MIT/Apache, actively maintained, implements PK Algorithm 2 — see `src/lib.rs` `add_dependency` → `dfs_forward`/`dfs_backward`/`reorder_nodes`). It's correct PK for edge insertion, but the API is fundamentally edge-oriented:

- `add_dependency` mutates internal `children`/`parents` sets, then attempts the reorder, then rolls back on `CycleDetected`. There's no "speculative check without commit" API.
- Our merge attempts run thousands of times per pop; each must be speculative (no rollback cost).
- Node contraction (the actual operation we need) would need to be simulated as a sequence of edge insertions, each going through the crate's commit/rollback dance.
- Using the crate would also require replacing `ClassId` with the crate's opaque `Node` handle.

Rolling our own kept the API contraction-shaped, kept the existing `ClassId`-indexed adjacency, and was ~250 lines.

### Performance

Measured on the existing `greedy_on_gaffer_chunk_completes_under_one_minute` benchmark (9709 owners, 23 proposals from factorize):

| Build mode | Baseline (cone-DFS) | This PR (PK)   | Delta |
|------------|---------------------|----------------|-------|
| fastbuild  | 320.24s             | 344.59s        | +7%   |
| opt        | 46.27s              | 46.63s         | +0.8% |

Both produce **identical output** (same 23 proposals, 8 extension). The fastbuild regression is debug-overhead noise; the opt timing is essentially equal.

**Why no big speedup on this fixture:** PK successfully removed the cycle check from the hotspot list (no longer appears in `perf report` flat output). But on the 9709-owner corpus, the cycle check wasn't the dominant cost to begin with — the bucketed perf profile in the linked perf doc shows BTreeMap walks (14%), iterators (7%), memmove (6%), etc. as the actual leaves. The 5-minute "large web corpus" the doc describes must be substantially bigger; PK's `O(|Δ|)` win materializes there.

The change is algorithmically correct and removes the worst-case `O(V²·d)` ceiling. On corpora where the cone-DFS DOES dominate, this will give the predicted 5–10× improvement.

### Verification

- All 73 tests in `//devinfra/js/debundle/...` pass, including:
  - `factorize_golden_output_unchanged` (byte-identical output)
  - `planner_and_materializer_agree_on_corpus`
  - `unification_byte_identical_on_well_formed_inputs`
  - `seed_skips_unrealizable_spec_module_contraction_and_reports` (exercises the `is_dag = false` fallback path)
- New `peel::topo_order::tests` cover: Kahn init on chains/diamonds, cycle check vs brute-force on random DAGs, contract sequences maintaining the invariant, transitive-predecessor reorders.

## Test plan

- [ ] `bazelisk test //devinfra/js/debundle/...` (73/73 pass locally)
- [ ] Byte-identical proposer output verified by `factorize_golden_output_unchanged`
- [ ] Larger corpus comparison (gaffer-private's full `modules propose` run) — left as a follow-up; the gaffer-private repo's CI would catch any output drift

## Follow-ups

- **Reverse index for `apply_contract` window scan**: currently iterates `self.ord` (`O(num_classes)`); a `pos_to_class: Vec<ClassId>` inverse would make it `O(|window|)`. Left out of this PR because the scan is cheap relative to other per-merge work.
- **Verify the larger-corpus win**: the perf doc's "large web corpus" is bigger than the test fixture I have. A follow-up benchmark on that corpus would confirm the predicted 5–10× speedup (or refute it, in which case the bottleneck is elsewhere — most likely the BTreeMap-backed adjacency, which is the next obvious target).

### DIFF
--- a/devinfra/js/debundle/BUILD.bazel
+++ b/devinfra/js/debundle/BUILD.bazel
@@ -322,6 +322,7 @@ rust_library(
         "peel/plan.rs",
         "peel/quotient.rs",
         "peel/test_utils.rs",
+        "peel/topo_order.rs",
     ],
     crate_root = "peel/mod.rs",
     edition = "2024",
--- a/devinfra/js/debundle/peel/mod.rs
+++ b/devinfra/js/debundle/peel/mod.rs
@@ -1,6 +1,7 @@
 pub mod factorize;
 pub mod plan;
 pub mod quotient;
+mod topo_order;
 
 pub use plan::{
     CommonArgs, ExplainArgs, ExplainReport, GraphSummaryArgs, GraphSummaryReport, OutputFormat,
--- a/devinfra/js/debundle/peel/quotient.rs
+++ b/devinfra/js/debundle/peel/quotient.rs
@@ -345,6 +345,15 @@ pub struct QuotientGraph {
     /// residual class transitions to non-residual via a contract
     /// that brings in a non-gate-residual member.
     next_module_idx: usize,
+    /// Incremental topological order over the live classes,
+    /// maintained across contractions via the Pearce-Kelly 2007
+    /// algorithm. Replaces the per-pop cone-DFS that was the
+    /// dominant inner cost — `merge_creates_new_constraining_cycle`
+    /// now reduces to an `O(|Δ|)` window-restricted reachability
+    /// check, and `apply_contract` runs a local Kahn over the
+    /// affected window `[ord[lo], ord[hi]]`. See
+    /// `peel/topo_order.rs` for the algorithm + invariants.
+    topo_ord: super::topo_order::TopoOrder,
 }
 
 /// One owner-edge with its `DepKind`-derived weight. Stored on the
@@ -374,15 +383,15 @@ fn edge_weight(kind: DepKind) -> u32 {
 /// present iff `weighted_count > 0`; the empty default is never
 /// stored.
 #[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
-struct EdgeState {
+pub(super) struct EdgeState {
     /// Number of underlying owner edges from members of the source
     /// class to members of the target class whose
     /// `constrains_init_order` is true. Was `class_edge_multiplicity`.
-    constraining_count: u32,
+    pub(super) constraining_count: u32,
     /// Total underlying owner edges (constraining or not) from
     /// members of the source class to members of the target class.
     /// Always `>= constraining_count`. Was `class_weighted_edge_count`.
-    weighted_count: u32,
+    pub(super) weighted_count: u32,
     /// Sum of `edge_weight` over all underlying owner edges. Was
     /// `class_edge_weight`.
     weighted_sum: u64,
@@ -526,9 +535,11 @@ impl QuotientGraph {
             realizability_index,
             class_module_id,
             next_module_idx,
+            topo_ord: super::topo_order::TopoOrder::empty(),
         };
         q.rebuild_class_adjacency();
         q.rebuild_cycle_cache();
+        q.rebuild_topo_ord();
         q
     }
 
@@ -606,6 +617,7 @@ impl QuotientGraph {
         q.rebuild_class_adjacency();
         q.rebuild_cycle_cache();
         q.rebuild_realizability_index();
+        q.rebuild_topo_ord();
         (q, group_class_ids)
     }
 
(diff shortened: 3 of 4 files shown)


agentydragon:
Follow-up commit `5bcc77279`: added the `pos_to_class: Vec<Option<ClassId>>` reverse index to `TopoOrder` and rewired `apply_contract` to slice it instead of scanning the full `ord` vector.

**Changes (1 file, +232 / -21):**
- `TopoOrder` now carries `pos_to_class` as the inverse of `ord` over live classes. Dead slots are `None`.
- `init_from_dag` builds `pos_to_class` in the same Kahn pass.
- `apply_contract` collects window classes from `pos_to_class[lo_ord..=hi_ord]` (`O(|window|)`) instead of scanning all `num_classes` slots. The Kahn re-rank inside writes to both `ord[c.0]` and `pos_to_class[new_rank]`; the trailing slot vacated by the loser is cleared.
- The empty-interior fast path (`lo_ord + 1 == hi_ord`) also updates `pos_to_class`.
- `validate()` asserts the inverse invariant in both directions (per live class and per live slot). It runs under `debug_assert!` after every `apply_contract` via `QuotientGraph`, so the assert covers the full debundle test corpus.

**New unit test:** `pos_to_class_is_inverse_of_ord_after_many_contracts` — random DAGs, long random safe-merge sequences, assert the inverse holds throughout. The invariant did not catch any bug during implementation; the asserts passed first try after the in-lockstep maintenance was wired up.

**Verification:**
- All 74 `//devinfra/js/debundle/...` tests pass (BuildBuddy invocation `b800d25d-1c3e-4f0f-bdd0-0ffb33dc84ca`).
- `modules propose` against the tana `78d928dca7` corpus emits **byte-identical JSON** (md5 `1fda9b1bab7bdf706cd4a63106a0554e`, 312 735 lines, both before and after).
- Opt-build wall time on the same corpus (3 runs each):

| run | before (`-c opt`) | after (`-c opt`) |
|----:|------------------:|-----------------:|
|   1 |           43.38 s |          43.47 s |
|   2 |           43.08 s |          43.13 s |
|   3 |           43.49 s |          43.19 s |

Within noise, as predicted in `perf/2026_05_27_propose.md` (the scan was ~500 ms total on the 9709-class fixture). The structural cost is now `O(|window|)`, so the worst case is gone.

**Branch state:** rebased onto `4b21c4a15` (current `origin/devel`); pushed with `--force-with-lease`.

PR #1706, opened 2026-05-27
debundle/realizability: cache base SCC partition for overlay verdict

agentydragon:
## Summary

- Cache the base graphs' SCC partitions on `IncrementalQuotient`
  (one `Vec<BTreeSet<ModuleId>>` plus a `ModuleId -> scc index` map
  for each of `constraining_graph` and `i_graph`). Invalidated
  together with the other base caches inside
  `invalidate_cached_simulator`, which is the single chokepoint
  already covering every base mutation site (`add_current_edge`,
  `remove_current_edge`, `rollback_graphs`).
- Add `OverlayGraphView::scc_containing_with_base_cache(module,
  cached_sccs, cached_scc_of)`. If `overlay.delta` is empty, return
  the cached SCC verbatim. If overlay edits exist but none of their
  endpoints sit inside `module`'s cached SCC, the overlay cannot
  perturb the partition for this query, so return the cached SCC.
  Otherwise fall through to the existing overlay-aware
  forward-intersect-reverse walk.
- Wire the cache into both `scc_containing` calls in
  `verdict_with_overlay_touching`.
- Gate per-call instrumentation (`AtomicUsize` counters +
  `AtomicU64` nanosecond accumulator) behind `DEBUNDLE_TIMING=1`;
  `SccTimingReporter::install_if_enabled` returns an RAII guard
  that prints the summary on drop.

## Verification

- `bazelisk test //devinfra/js/debundle/...:all`: 75 tests pass.
- `modules propose --format json` against the tana fixture is md5
  byte-identical to the devel baseline at `9c5c369c6`
  (`1fda9b1bab7bdf706cd4a63106a0554e`).

## Wall-time measurement

Tana fixture, 5 alternating rounds (devel `9c5c369c6` vs this
branch), both binaries built `-c opt --extra_rustc_flag=-Cdebuginfo=1`,
single-threaded:

| round | baseline wall | new wall | full walks | cache hits |
|------:|--------------:|---------:|-----------:|-----------:|
|     1 |        45.41s |   47.01s |       4380 |          0 |
|     2 |        46.70s |   46.44s |       4380 |          0 |
|     3 |        46.43s |   45.13s |       4380 |          0 |
|     4 |        43.75s |   43.96s |       4380 |          0 |
|     5 |        45.84s |   45.63s |       4380 |          0 |

Cache-hit rate is zero on tana: every overlay edit has at least one
endpoint inside the queried module's base SCC. This is structural --
`verdict_after_moving_owners_touching` always queries `to` (the
candidate destination of the move), and every overlay edge is by
construction incident to a moved owner, whose post-move module is
`to`. So the heuristic short-circuit cannot fire on this workload
and wall time is unchanged within noise. The infrastructure remains
as the foundation for follow-on optimizations (sparser overlays, or
queries against modules other than the move target).

## Test plan

- [x] `bazelisk test //devinfra/js/debundle/...:all` — 75 pass.
- [x] `modules propose --format json` against tana md5-identical to baseline.
- [x] 5 alternating wall-time rounds reported above.

### DIFF
--- a/devinfra/js/debundle/BUILD.bazel
+++ b/devinfra/js/debundle/BUILD.bazel
@@ -595,6 +595,7 @@ rust_binary(
     crate_root = "main.rs",
     edition = "2024",
     deps = [
+        ":analysis",
         ":cli",
         "@crates//:anyhow",
         "@crates//:clap",
--- a/devinfra/js/debundle/lib.rs
+++ b/devinfra/js/debundle/lib.rs
@@ -62,7 +62,7 @@ pub use purity::{
 };
 pub use realizability::{
     CrossRebindEdge, DeltaHandle, PartitionDelta, RealizabilityIndex, RealizabilityVerdict,
-    SccDiagnosis, check_realizability, check_realizability_with_quotient,
+    SccDiagnosis, SccTimingReporter, check_realizability, check_realizability_with_quotient,
 };
 pub use reports::schema::{
     AtomicGraphReport, AtomicUnitEdgeReport, AtomicUnitReport, BindingReport, EdgeRoleReport,
--- a/devinfra/js/debundle/main.rs
+++ b/devinfra/js/debundle/main.rs
@@ -1,10 +1,15 @@
 use std::process::ExitCode;
 
+use analysis::SccTimingReporter;
 use clap::Parser;
 use debundle_cli::{DebundleArgs, run_debundle_cli};
 use swc_common::{GLOBALS, Globals};
 
 fn main() -> ExitCode {
+    // Install the realizability scc_containing timing reporter when
+    // DEBUNDLE_TIMING is set. The guard prints a summary on drop.
+    let _scc_timing_guard = SccTimingReporter::install_if_enabled();
+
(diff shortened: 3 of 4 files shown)


agentydragon:
Closing as a measured dead end on this workload. The cache infrastructure is correct but the heuristic short-circuit ("if overlay doesn't touch queried SCC, return cached") gets 0/4380 cache hits on tana: verdict_after_moving_owners_touching always queries the move destination to, and every overlay edge is by construction incident to a moved owner whose post-move module is to — so the overlay always intersects to's SCC and the short-circuit always falls through.

The instrumented finding IS valuable: scc_containing accounts for ~22% of total wall (4380 calls × ~2.3 ms each = ~10s of 46s). Real fix is incremental SCC maintenance via Pearce 2003 (companion to the topological-order PK we already implemented in peel/topo_order.rs). See perf/proposer_roadmap.md P1 #1 for the spec.

Branch debundle-cache-base-scc preserved for reference but should not merge — it adds ~240 lines of cache machinery for zero measurable benefit on this workload.

PR #1736, opened 2026-05-30
airlock: reconcile managed-secret annotations every refresh pass

agentydragon:
Adds `K8sTokenStore.reconcile_annotations()` and calls it from the token refresh loop for every provider's refresh + access secret on every pass — even when the token doesn't change.

## Why

airlock only (re)writes a managed token secret when it writes a token. For providers whose tokens never refresh (e.g. Plaid access tokens, which never expire), a config-level annotation change — like adding a reflector target namespace to an `access_secret` — never reached the live secret without a manual re-link. Now it propagates on the next refresh-loop pass.

## What

- `reconcile_annotations(secret, ns, annotations)` merges the configured annotations onto an existing managed secret (adds/updates the configured keys, leaves any others; no-op when already in sync or when the secret doesn't exist yet). Does not touch token data.
- The refresh loop reconciles annotations for each provider's refresh + access secret every pass (runs on startup and every `check_interval`), independent of `needs_refresh`.

## Tests

- `airlock/oauth/test_k8s_client.py` (new): patch-on-drift / no-op-when-in-sync / no-op-when-absent / no-op-when-no-annotations-configured.
- `airlock/oauth/test_refresh.py`: reconciliation runs for a fresh, non-refreshed token.

`bbr test //airlock/oauth/...` green.

This is split out of the in-progress Plaid MCP work so it can be reviewed/merged independently.

https://claude.ai/code/session_01D4nPp3SfsTPgnV3NmcnZvs

---
_Generated by [Claude Code](https://claude.ai/code/session_01D4nPp3SfsTPgnV3NmcnZvs)_

### DIFF
--- a/airlock/oauth/BUILD.bazel
+++ b/airlock/oauth/BUILD.bazel
@@ -7,7 +7,7 @@ py_library(
     srcs = ["provider.py"],
     visibility = [
         "//airlock:__subpackages__",
-        "//x/plaid:__pkg__",
+        "//plaid:__pkg__",
     ],
     deps = [
         "@pypi//httpx",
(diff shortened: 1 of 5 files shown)


agentydragon:
Superseded by a same-repo PR off `devel`: fork PRs don't receive `SOPS_AGE_KEY`, so bazel-ci can't decrypt SOPS files and fails at setup. Reopening from an `origin` branch where CI has secrets. (Change is unchanged and green via `bbr test //airlock/oauth/...`.)

---

PR #1735, opened 2026-05-30
plaid_utils: read-only Plaid MCP server on the official plaid-python SDK

agentydragon:
Read-only MCP server over my Plaid-linked bank accounts (Chase + BofA): transactions,
balances, and liabilities (credit cards, mortgages, student loans). Standalone
Authentik-gated endpoint at `plaid-mcp.allegedly.works`, fronted by the shared
`mcp-oauth-facade` (the `manifold-mcp` two-container pattern); the Python server itself is
auth-oblivious.

Rebased onto current `devel`. The airlock-annotation-reconcile commit that was on this
branch is dropped — it landed independently as #1737 and was then superseded by the ESO
switch in #1738.

### `plaid_utils/` — client + MCP server

- **Official SDK, not hand-rolled.** `client.py` is a thin async wrapper over
  [`plaid-python`](https://github.com/plaid/plaid-python) (v39.2.0). The SDK is
  synchronous, so calls run in threads via `asyncio.to_thread`; responses go through the
  SDK's `sanitize_for_serialization` and are validated into typed `plaid_utils.models`
  Pydantic models at the boundary — callers get typed objects, not raw dicts.
- Package is `plaid_utils/` (not `plaid/`) because the SDK owns the top-level `plaid`
  import; a local `plaid` package would collide. Mirrors the existing `openai_utils`
  convention.
- **5 bounded, paginated tools** (`mcp_server/server.py`): `list_items`, `list_accounts`
  (cached balances), `list_transactions` (date range + `offset`/`count`,
  `total`-before-slicing), `get_liabilities` (credit/mortgage/student, each null when
  absent; gated on the `liabilities` product), `get_live_balance` (real-time, rate-limited).
  API warts (amount sign, `/transactions/get` vs `/sync`, pending→posted,
  `ITEM_LOGIN_REQUIRED` → re-link via airlock) documented in the tool descriptions.

### Cluster deploy (`cluster/k8s/agents/plaid-mcp/`)

- Two-container pod (auth-oblivious server `:8080` + `mcp-oauth-facade` `:8765`), HTTPRoute,
  valkey for OAuth state, CiliumNetworkPolicy (ingress→`:8765` only), no ServiceAccount
  (`automountServiceAccountToken: false`).
- Authentik provider/application/group + `plaid-mcp-oidc` secret via the
  `agent-machine-access` Terraform module.
- **Secrets via ESO** (not the reflector): three `ExternalSecret`s mirror
[...]

PR #1745, opened 2026-05-30
cluster/validation: image-automation check via build output, not raw YAML walk

agentydragon:
## The bug

`check_image_automation_webhook` (added in #1741) `rglob`-walked every `cluster/k8s/*.yaml`
and re-parsed it with `yaml.safe_load_all`. That:

- **crashed** on the Authentik blueprints' custom `!Find`/`!Env` tags
  (`ConstructorError: could not determine a constructor for the tag '!Find'`), taking down
  the whole `ducktape-precommit` cluster-validate hook;
- bypassed the parsed-manifest pipeline every other check uses, re-reading the working tree
  directly (so it also saw unstaged files); and
- was only ever exercised by **synthetic unit fixtures**, never the real tree — which is how
  the crash shipped to the pinned `ducktape-git-hooks` wheel and started failing pre-commit
  on every PR that touches `cluster/k8s`.

## The fix

Rewrite it to consume `ParsedCluster.build_results` — the rendered kustomize/flux output the
validator already produces — like the other checks. `KustomizeBuildResult` now also keeps the
raw rendered `docs`, so checks needing non-`HelmRelease` spec fields
(`ImagePolicy.spec.imageRepositoryRef`, `Receiver.spec.resources`) can read them via small
typed models. Rendered blueprints are ConfigMap **string data**, so their custom tags never
reach the YAML parser — no crash, and no source re-walk.

Add `test_cluster_integration::test_image_automation_webhook_consistency`, which runs the check
against the **real built cluster** — closing the coverage gap that hid the bug.

## Verified

`bbr build //cluster/validation/...` (lint/mypy) + `bbr test //cluster/validation/...` — 10/10
PASSED, including the new real-tree integration test.

## ⚠️ Pre-commit will be red on this PR

The `ducktape-precommit` hook runs from the **pinned wheel** (`nix/artifact-pins.json`), which
`bc40776` (`ducktape-automation[bot]`) just advanced to the #1741-inclusive (buggy) build —
so `devel`'s own pre-commit is currently red, and so is every PR's `--all-files` run. This
source fix only takes effect once the wheel is rebuilt from it and the pin advances. Options to
un-break immediately (e.g. reverting the `ducktape-git-hooks` pin to the prior good wheel so the
post-merge release+sync-pins moves forward to the fixed one) — happy to add that here if wanted.

https://claude.ai/code/session_01D4nPp3SfsTPgnV3NmcnZvs

---
_Generated by [Claude Code](https://claude.ai/code/session_01D4nPp3SfsTPgnV3NmcnZvs)_

### DIFF
--- a/cluster/validation/image_automation.py
+++ b/cluster/validation/image_automation.py
@@ -6,53 +6,16 @@ picked up on the 5-minute `ImageRepository` poll. Also checks that every `ImageP
 references a defined `ImageRepository`, and that the webhook doesn't reference one that no
 longer exists.
 
-Operates on the kustomize build output (`ParsedCluster.build_results`) — the rendered
-manifests the rest of the validator already produces — rather than re-walking source YAML.
-That keeps it consistent with the other checks, sees exactly what Flux applies, and avoids
-choking on the Authentik blueprints' custom `!Find`/`!Env` tags (which only appear as
-ConfigMap string data once rendered, never as document-level YAML tags).
+Reads the typed resources from `ParsedCluster.build_results` — the kustomize/flux build
+output the validator already produces — and isinstance-dispatches on the parsed variants, so
+it sees exactly what Flux applies and never re-walks source YAML (which would choke on the
+Authentik blueprints' custom `!Find`/`!Env` tags).
 """
 
 from __future__ import annotations
 
-from pydantic import BaseModel, ConfigDict, Field
-from pydantic.alias_generators import to_camel
-
 from cluster.validation.cluster import ParsedCluster
-
-
-class _Meta(BaseModel):
-    model_config = ConfigDict(extra="ignore")
-    name: str = ""
-
-
-class _Manifest(BaseModel):
-    """Just enough of a rendered manifest to route it by kind."""
-
-    model_config = ConfigDict(extra="ignore")
-    kind: str = ""
-    metadata: _Meta = Field(default_factory=_Meta)
-
-
-class _RepoRef(BaseModel):
-    model_config = ConfigDict(extra="ignore")
-    name: str = ""
-
-
-class _ImagePolicySpec(BaseModel):
-    model_config = ConfigDict(extra="ignore", populate_by_name=True, alias_generator=to_camel)
-    image_repository_ref: _RepoRef = Field(default_factory=_RepoRef)
-
-
-class _ResourceRef(BaseModel):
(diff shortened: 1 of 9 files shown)


cluster/validation/image_automation.py
```diff
@@ -5,44 +5,72 @@
...
+    for result in cluster.build_results:
+        for doc in result.docs:
+            manifest = _Manifest.model_validate(doc)
+            spec = doc.get("spec") or {}
+            if manifest.kind == "ImageRepository":
+                image_repos.add(manifest.metadata.name)
+            elif manifest.kind == "ImagePolicy":
+                policy_refs[manifest.metadata.name] = _ImagePolicySpec.model_validate(spec).image_repository_ref.name
+            elif manifest.kind == "Receiver":
+                receiver = _ReceiverSpec.model_validate(spec)
```
agentydragon: Maybe use pydantic discriminated union parsing?

cluster/validation/kustomize.py
```diff
@@ -79,6 +80,11 @@ class KustomizeBuildResult(BaseModel):
 
     kustomization_path: Path
     resources: list[K8sResource] = []
+    # Full rendered manifests as raw dicts. `resources` is the typed,
+    # HelmRelease-focused narrowing of these; checks that need other kinds' spec
```
agentydragon: If you just add the manifests you care about into the common pool you might not need to carry the raw dicts

cluster/validation/image_automation.py
agentydragon: Done in d7827eb — `parse_k8s_resources` is now a kind-discriminated parser: it returns the typed subclass per kind (`HelmReleaseResource` / `ImageRepositoryResource` / `ImagePolicyResource` / `ReceiverResource`), and the generic `K8sResource` base for everything else. The check isinstance-dispatches over those variants (no kind-string compares, mypy-narrowed).

---

cluster/validation/kustomize.py
agentydragon: Done in d7827eb — those typed variants now land in the common pool (`build_results.resources`), so the `docs` field on `KustomizeBuildResult` is gone; the check reads the typed resources directly. Also rebased onto current devel (with #1735 merged) — its plaid-mcp-server ImageRepository is now exercised by the real-tree `test_cluster_integration` check, still consistent.

---

PR #1815, opened 2026-06-02
augur(frontend): compare multiple scenarios in the product view

agentydragon:
## What

Lets the product projection view hold a **set** of scenarios and compare their distributions side by side — e.g. "what if I took a mortgage instead of renting". Reintroduces the multi-scenario comparison called for in `plans/roadmap.md` and `sim/TODO.md`.

- **Scenario bar** — add / select / rename (double-click) / delete scenarios, capped at 5, one color per scenario.
- **Overlaid fan chart** — every scenario draws as a median + P5–P95 band in its own color, with a legend and a per-scenario median tooltip. To keep it readable with several scenarios, the inner P25–P75 band and edge lines are dropped in multi-scenario mode. A lone scenario renders **byte-identically** to before (inner band + edges in blue), so single-scenario chart goldens are unchanged apart from the new bar above them.
- **Terminal scenario comparison table** — one column per scenario (median terminal per metric, P5–P95 beneath), computed client-side from each scenario's rollout summaries.
- The rollout **histogram, selected-rollout overlay, events, and the detailed terminal-percentile table** scope to the active scenario (`sim/TODO.md`'s "selection scopes to one scenario at a time").

## No backend change

Every scenario shares the same seed set, and identical seeds reproduce identical sampled exogenous paths (each factor draws from its own `np.random.default_rng(seed)`), so the overlaid fans are apples-to-apples without a comparison endpoint. The frontend just fans out one existing `/api/product/projections/metric_fan` request per scenario; results land per-scenario in place so comparison fans don't blank out while the active scenario is edited.

## URL state

The set rides a `?scenarios=` JSON param. A single scenario keeps the compact, shareable `?s=`/`?lc=` form, and **any pre-multi link decodes as a 1-element set** (backward compatible). A malformed or unrecognized-version `?scenarios=` falls back to a single default scenario.

## Tests

- New `//augur/frontend:input_helpers_test` — vitest round-trips for the scenario-set codec (single ↔ `?s=`, multi ↔ `?scenarios=`, legacy backward-compat, lifecycle-event preservation with regenerated keys, malformed/version fallbacks).
- New `product_scenario_comparison` visual golden (rent vs. mortgage); the two existing product goldens were regenerated for the new scenario bar (chart pixels unchanged); `calibration_page` is byte-identical.
- `tsc`, type-aware `eslint`, both vitest suites, and the visual test pass on RBE.

`BAZEL_TEST_INVOCATIONS=buildbuddy:1b1c771a-23a1-4863-9e08-8e98d01411f4,buildbuddy:d2898c75-4255-4a71-adac-0ca534652937`

## Follow-ups (left as TODOs)

- Side-by-side editing of several scenarios' controls at once (today you select a scenario via the chip bar and edit its form).
- The delta / paired-difference distribution view (`TODO.md`) is unchanged — genuinely separate future work.

https://claude.ai/code/session_01EcdxgZg2XD5jg4Eef9hqvg

---
_Generated by [Claude Code](https://claude.ai/code/session_01EcdxgZg2XD5jg4Eef9hqvg)_


agentydragon:
Superseded by #1817, which has this same work plus the design-mockup renderer (it's pushed from the `agentydragon/ducktape` branch rather than the fork). Closing in favor of that one.

---

PR #1867, opened 2026-06-04
augur/plans: JAX migration plan for the simulation numeric core

agentydragon:
Adds `augur/plans/jax_migration.md` — the design/plan for a full migration of augur's forward-simulation numeric core (exogenous sampling + the dense month loop) from imperative NumPy to JAX (`jit` + `lax.scan` + `vmap`), keeping `compile_simulation` and the polars decode on the host.

## Highlights

- **Why now**: post-optimization, `metric_fan` (10k × 100mo) is ~11.6s and entirely simulation/sampling-bound. JAX targets the two real levers: the **correct per-seed-independent vectorized RNG** (`random.fold_in` + `vmap` — the thing NumPy can't do natively, which blocked the GBM vectorization) and a **fused month loop** (`lax.scan`).
- **Amdahl honesty**: the traceable core is ~8s of ~11.6s; `compile` (~2.1s) and polars decode (~2.4s) stay host → a **~4.5s floor** on CPU unless those move on-device too. Single-process CPU win ≈ 2.5×; the big wins are GPU at large R.
- **Feasibility, grounded in the code**: state buffers are already **static-shaped** `(snapshots, count, R)` and lot disposition/settlement are **mask-based (no dynamic-shape FIFO/searchsorted)** — the key "is it possible" risk is clear. The cost is the functional rewrite of `phases.py` (~132 control-flow + ~59 masked ops); the **tax machinery is the hardest** and goes last.
- **Phased plan** (each gated by a NumPy-vs-JAX parity harness): **0** infra + `jaxlib` under Bazel/NixOS/RBE + x64 + parity helper; **1** sampling (proving ground, fixes the RNG); **2** month loop via `lax.scan` (the bulk); **3/4** optional on-device decode + GPU.
- **Risks/decisions**: total golden churn, `jaxlib` build integration, x64 vs float32, jit compile latency, debuggability, dual-impl during migration; decisions to settle before Phase 0.

Docs-only (`augur/plans/`).

https://claude.ai/code/session_01VYv5maUG91A17axRLcNNMZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01VYv5maUG91A17axRLcNNMZ)_

PR #1869, opened 2026-06-04
augur/model: JAX migration slice 1 — backend switch + dual-impl GBM

agentydragon:
First executable step of the JAX migration (see #1867 / `augur/plans/jax_migration.md`): the **parallel-implementation + shared-test machinery**, proven on the GBM sampler.

## What

- **`sim_backend.py`** — the switch. `current_backend()` selects NumPy (reference, default) vs JAX; `use_backend(...)` scopes it. A `ContextVar` (thread-safe — the API server is concurrent), no module global. A leaf module both the model layer and the sim engine can read.
- **`gbm.py`** — dual implementation behind the switch:
  - `_sample_levels_numpy` — the existing per-rollout `default_rng(seed)` loop (unchanged, the reference).
  - `_sample_levels_jax` — a `vmap` over **per-seed PRNG keys** (each rollout's key folded from its own seed in 32-bit words). This is the "vector of R independent seeded states": trajectory `i` depends only on `rollout_seeds[i]`, never on the batch.
- **`gbm_test.py`** — one suite **parametrized over both backends** asserting the same invariants (the two backends produce different *realized* values by design, so these are properties, not samples):
  - per-seed **reproducibility**;
  - **per-seed independence** — a seed's path is identical sampled alone or in a batch (exactly what a single batch-seeded draw would break);
  - **deterministic** parity at `sigma=0`;
  - the configured **log-return moments**.

## Notes

- **No new dependency** — JAX already builds in-repo (used across `augur/model` + `augur/fit` via `numpyro`), so the heavy Phase-0 build risk was already retired.
- The JAX path runs at float32 (repo default; no `jax_enable_x64`). That's fine here because GBM is RNG (values change by design); the **x64 decision is deferred to the deterministic-engine slices**, where exact NumPy-vs-JAX parity is the oracle.

## Test

`//augur/model:gbm_test` passes against **both** backends; `//augur/model:series_model_test` unaffected.

This establishes the pattern every subsequent slice (other samplers, then the engine phases) will follow.

https://claude.ai/code/session_01VYv5maUG91A17axRLcNNMZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01VYv5maUG91A17axRLcNNMZ)_


agentydragon:
Closing as superseded. The full always-scan JAX engine landed in #1905, which absorbed this slice's entire deliverable — `augur/model/sim_backend.py`, the dual-impl `gbm.py` (`_sample_levels_jax` / `use_backend`), the backend-parametrized `gbm_test.py`, and the conftest backend fixture are all on `devel` now. The per-call recompile that the engine had on top of that is fixed in #1912.

---

PR #1887, opened 2026-06-04
augur: fix bayesian_mint_streams calibration 500 — clamp public-market CDF to the horizon

agentydragon:
## What

Root-cause fix for the `/api/calibration/run` **500 (`IndexError: index 2 is out of bounds for axis 0 with size 2`)** seen with the `bayesian_mint_streams` preset.

`_public_market_marginal_cdf` (`augur/model/private_equity_risk.py`) allocates `cdf` of length `horizon_months + 1`, but its anchor loop wrote `cdf[t]` up to each `public_market_cdf_anchor.month`. When an issuer's IPO anchor month exceeds the sampling horizon, the write runs off the end. Scrolling the Calibration tab's chart wheel down to a short horizon drove the sample at `horizon_months=1` → `cdf` size 2 → `cdf[2]` overflow. Only `bayesian_mint_streams` hit it (the only preset whose PE issuer carries explicit `public_market_cdf_anchors` routed through the rounds path).

## Fix

Clamp the write loop to `min(anchor.month, horizon_months) + 1`; `prev_month`/`prev_cum` still advance to the true anchor so the flat tail and subsequent anchors stay anchored correctly. Now **no** caller can overflow it — calibration at any horizon, a direct API request, or any future short-horizon sampler.

Regression test (`private_equity_risk_test.py`): samples a `public_market_cdf_anchors` issuer at a horizon between two anchors (12 / 36, horizon 24) and asserts no raise + a monotone CDF of length `horizon+1`.

## Note on the UI

The Calibration tab's UI symptom is independently handled by #1885 (already on devel), which pins calibration sampling to the deployment max horizon and removes the wheel zoom from the calibration fans — so the tab no longer issues short-horizon runs. This PR is the **defense-in-depth root-cause fix** so the function is safe regardless of caller.

## Test plan

- [x] `bbr build //augur/...` — clean (lint + mypy)
- [x] `bbr test //augur/model:private_equity_risk_test` — pass (incl. the new regression test)

https://claude.ai/code/session_014ZqWEw2nP8z4vajVovtjE4

---
_Generated by [Claude Code](https://claude.ai/code/session_014ZqWEw2nP8z4vajVovtjE4)_

### DIFF
--- a/augur/model/private_equity_risk.py
+++ b/augur/model/private_equity_risk.py
@@ -1105,7 +1105,10 @@ def _public_market_marginal_cdf(issuer: PrivateEquityRiskIssuerConfig, *, horizo
         survival_ratio = (1.0 - anchor.cumulative_probability) / max(1.0 - prev_cum, 1e-12)
         # Per-month survival factor that interpolates the anchor CDF exactly.
         per_month_survival = survival_ratio ** (1.0 / span)
-        for t in range(prev_month + 1, anchor.month + 1):
+        # Clamp the write loop to the horizon: an anchor month beyond `horizon_months` would
+        # index past `cdf` (length `horizon_months + 1`). `prev_month`/`prev_cum` still advance to
+        # the true anchor below so the flat-tail and later anchors stay anchored correctly.
+        for t in range(prev_month + 1, min(anchor.month, horizon_months) + 1):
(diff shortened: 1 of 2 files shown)


agentydragon:
Superseded by #1888, which landed the identical clamp (`min(anchor.month, horizon_months) + 1`) on devel first. Closing as a duplicate — the root-cause fix is already in.

---

PR #1929, opened 2026-06-08
augur: fix terminal-distribution plot never rendering + rollout 422 on negative liquid net worth

agentydragon:
## Summary

The Augur product page's terminal-distribution plot was broken, failing `//augur:visual_test` (all 7 product goldens) and `//augur:browser_shell_test` on RBE/CI. devel has been red on these for a while. Two independent root causes:

### 1. `svgWidth` measurement race — plot never renders (`terminal_distribution.tsx`)

The SVG width was measured in a `useLayoutEffect([])` that read `containerRef.current`. But the measured container is gated behind `if (series.length === 0) return null`, so it only mounts once results have loaded. The one-shot effect runs after the **first** commit — typically before results arrive, when the node doesn't exist yet — hits `if (!container) return`, and because deps are `[]` it **never runs again**. So `svgWidth` stays `null`, only the header renders, the plot SVG (`[data-product-terminal-distribution-plot]`) never appears, and the density wait times out.

This is why it was intermittent: a fast first paint (data already present) won the race and rendered; a reload or a slow CI worker lost it. Diagnosed from the live DOM: `hasContainer: true, hasPlot: false, seriesCount: 0` — i.e. the `svgWidth == null` branch, stuck.

**Fix:** replace the one-shot effect with a **callback ref** that measures (and attaches a `ResizeObserver`) whenever the node actually attaches, however late.

### 2. Rollout `422` on negative liquid net worth (`wire.py`)

With the plot rendering again, clicking it selects a rollout and fetches `/api/product/projections/rollout` — which returned **422** for the property-lifecycle scenario, so no selected-rollout line appeared. The 422 was a **response**-model validation error:

```
loc: ["liquid_net_worth_usd"], input: -22182.87, "Input should be greater than or equal to 0"
```

`PrivateEquityOpportunityEvent.liquid_net_worth_usd` was `NonNegativeFloat`, but liquid net worth at a PE-tender opportunity can legitimately go negative when spending/obligations outrun liquid assets. FastAPI's `response_model` validation then rejected the entire rollout response.

**Fix:** type it `float` (matching `TerminalMetrics.liquid_net_worth_usd`, which is already `float`).

## Goldens

Regenerated the 7 product visual goldens for the now-correctly-rendered terminal distribution (`UPDATE_GOLDEN=1`).

## Verification

On RBE: `//augur:visual_test` and `//augur:browser_shell_test` both pass **3/3** `--runs_per_test` (the original failure was a race, so determinism was checked explicitly). Lint clean.


agentydragon:
Superseded by #1931 — same two fixes rebased onto current devel (4d5af7db2) with freshly regenerated goldens; #1929's base and goldens went stale behind the #1924–#1928 augur/sim refactors.

PR #1968, opened 2026-06-10
augur evidence scraper: skip refetch when committed evidence is still fresh

agentydragon:
## What

Adds a freshness gate to the evidence scraper's `run_scrape`: if `HEAD` (the last committed refresh) is younger than `--max-age-hours` (default **20h**), skip the whole refresh and exit 0.

Rationale: when the committed evidence is still fresh, re-GETting all 11 upstreams just re-downloads unchanged data (the 92 MB Zillow CSV, etc.) and burns upstream rate limits — we've observed FRED `504`s under repeated hits.

## Freshness signal

Per the chosen approach: **git commit age**. Since every run commits all sources in one commit, `HEAD`'s age *is* each file's last-change age — so this is effectively the per-file commit-age check, computed without history (the scraper shallow-clones at `depth=1`, so no per-file history is available).

Behavior:
- daily CronJob (runs >20h apart) → `HEAD` is >20h old → still refreshes every day;
- closely-spaced re-runs (manual triggers, retries within 20h) → skip, stop re-hammering upstreams.

## Tests

`test_run_scrape_skips_when_evidence_is_fresh` (HEAD 1h old → no fetch, no commit) and `test_run_scrape_refetches_when_evidence_is_stale` (HEAD 48h old → fetches + commits). 9/9 pass.

## Related follow-up (not in this PR)

Verified live: the pipeline now works end-to-end (repo fully populated, git-sync serving `/git/evidence`). But a transient single-upstream outage (e.g. a FRED `504`) currently makes the whole Job exit 1 → red + backoff, even though it commits the other 10 sources fine. Worth softening the exit-code contract (fail only when a source that's actually *stale* can't be refreshed) — happy to do that as a separate change.

### DIFF
--- a/finance/augur/ingest/fetch.py
+++ b/finance/augur/ingest/fetch.py
@@ -20,7 +20,7 @@ import os
 import sys
 import tempfile
 from collections.abc import Iterable
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
 import pygit2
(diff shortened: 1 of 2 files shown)


agentydragon:
Superseded by the per-source last-fetch manifest approach (new PR) — global HEAD-age gate couldn't isolate a stuck source.

PR #2064, opened 2026-06-11
debundle: close soundness holes in fact extraction and the S-chain

agentydragon:
Closes a set of under-restriction soundness holes in the debundler's fact extraction (`facts/mod.rs`) and side-effect chain (`graph.rs`): shapes where the validator accepted a spec whose emitted bundle misbehaves at runtime. Every hole landed with a minimal e2e fixture verified red against the pre-fix tree (BuildBuddy invocation `21e63e08-2390-41e3-8cc3-a314aec3c25d` shows all five new test targets failing pre-fix; the two reshaped hoisted-`var` fixtures were re-verified red via stash in `redcheck` runs).

### 1. At-init promotion: unresolvable callees silently skipped

**Counterexample (gate-accepted, runtime TDZ):** mod_a: `const A = 1; function readB(){return B;} const g = readB; const r = g();` / mod_b: `const B = A + 1;` — asymmetric I-cycle, the simulator accepted it, `g()` threw `ReferenceError` under node. Method-call variant: `const api = { read: () => B }; api.read();`. `at_init_calls` recorded only bare-`Ident` callees (so `api.read()` was invisible), and promotion `continue`d on callees that didn't resolve to a chunk function (so the alias `g` was skipped). design.md's claimed "validator's strict rule" safety net did not exist.

**Fix (task option (a)+(b) hybrid):** a callee is *precisely resolvable* only when its binding is declared directly as a function value (`function f`, single-declarator `const f = () => …`) and never rebound. Everything else takes a conservative fallback: the statement eagerly depends on the full lazy closures of every owner reachable (through the chunk's read graph) from the bindings the unresolved call mentions — callee root, arguments, computed keys — plus the statement's own closures when the call carries an inline function (IIFE, `arr.forEach(x => …)`). The fallback propagates transitively through the promotion call graph (a resolvable callee whose body contains an unresolved call escalates its at-init callers). Two deliberate scoping decisions that need a maintainer eye:

- **Fallback rebind targets get order-only `EagerUse` edges**, not `EagerRebind`: write legality is already enforced by the direct `LazyRebind`/`DeferredRebind` edge from the statement lexically containing the write; emitting `EagerRebind` from the *caller* forced the call site into the co-location unit and rejected realizable pinned shapes (`accepted_spec_runs_under_node_test::peeled_method_reassigning_top_level_let`).
- **Fallback edges sourced in residual are dropped at partition projection** (`partition_endpoints`): residual is the ESM DFS root and evaluates last, so an at-init call from residual code cannot observe a TDZ — and the emitter emits no entry-side phantom imports, so keeping these edges would feed the Pass-2 simulator an I-topology the emitted bundle doesn't have (it flipped `mediator_reaches_asymmetric_cycle_test` from reject to accept while the bundle still TDZ'd). Marked by the `PromotedAtInit { callee_owner == from }` convention.

Residual unmodelled channels (global stashes, rebound bindings, function parameters, `new`, cross-chunk) are now documented as preconditions in design.md's Limitations, replacing the dangling claim.

Tests: `e2e/at_init_unresolved_callee_test.rs` (alias + method-call rejections, acyclic green companion).

### 2. Nested-closure rebinds never reached cross-rebind rejection

Only `first_order_lazy_rebinds` emitted edges; a rebind two closures deep (`globalThis.__updateState = () => { state = "updated" }`) emitted nothing, so the split spec was accepted and the emitted bundle threw `TypeError` the moment the stored closure ran — the ESM read-only-imports rule is time-independent. New `DepKind::DeferredRebind` participates in cross-destination-rebind rejection (`is_rebind()`) and bidirectional `G_atomic` co-location, but **not** in init-order constraints (`constrains_init_order() == false`) or the I-graph — preserving the original purpose of the nested-closure narrowing (no manufactured init-order constraints; post-await test still green).

Tests: reworked `e2e/at_init_promotion_nested_closure_test.rs` — split spec now rejected; co-located spec green with a post-init `assert_generated_module_after_entry_script` probe observing the update through the live export binding; the init-order contract retained via a nested-*read* fixture with a post-init probe.

### 3. Block-hoisted `var` declarations were invisible

`try { var impl = "primary"; } catch (e) { var impl = "fallback"; }` declared `impl` invisibly: readers got no owner edge and no import wiring (runtime `ReferenceError`), and `try { var Math = shim; } catch {}` defeated the A8 shadowed-globals computation (`Math.PI` still classified as the pristine whitelisted global). Fix: shared `binding_targets::hoisted_var_ids` (function/arrow/accessor/class boundaries) feeds `collect_declared_names`, the shadowed-globals set, and the lowering side's `declaration_by_name`. Two emission-side consequences: bindings declared by anonymously-claimed statements now follow the claiming module (`adopt_bindings_of_claimed_anonymous_statements` — ownership, exports, cross-module imports), and the auto-grow pass only grows entry exports for names the final entry body actually declares (previously it could emit `export { impl }` with no backing declaration — SyntaxError at load). The `compute_shadowed_globals` edit is just the hoist-aware feed via `collect_declared_names`, per the parallel-branch constraint.

Tests: `e2e/hoisted_var_declaration_test.rs` (split reader wired end-to-end; co-located green; `Math.PI` shim variant rejected).

### 4. `export default` purity

`classify_item` maps non-`ExportDecl` module decls to `Export` and `item_purity` mapped `Export → Pure`, so `export default sideEffect()` and `export default class C { static { … } }` dropped out of the S-chain and other modules' statements reordered around them. `ExportDefaultExpr` now routes through `classify_expr_purity`, `ExportDefaultDecl(Class)` through `class_has_static_observable`.

Tests: `e2e/export_default_purity_test.rs` (both shapes rejected when the restored edge is unsatisfiable; green companion pinning preserved `SCZ` ordering).

### 5. Write-after-read edges

Last-writer-only tracking missed WAR: `globalThis.flag = "first"; const snapshot = globalThis.flag; globalThis.flag = "second";` let the second writer schedule before the reader. `emit_s_chain` now tracks `readers_since_last_write` per cell; a writer sequences after every reader since the cell's last write.

### 6. Opaque calls

Any at-init call / `new` / optional call / tagged template the purity classifier does **not** prove `Pure` flips `dataflow_summarizable = false` (policy-phase check, so classifier-Pure calls — whitelists, `pure`/`pure_new` annotations, recursively-pure chunk functions — stay summarizable). This covers `console.log` pairs (I/O isn't a cell), `function f(){globalThis.x=1} f();` (the depth-0-gated cell recorder can't see callee bodies), and subsumes indirect `(0, eval)(…)`. This deliberately degrades the optimization for call-heavy code — documented honestly in README. The pinned constructor-call test was updated accordingly (`unproven_constructor_calls_keep_conservative_s_edge`): `new <chunk class>` is `UnknownNew` without a `pure_new` annotation, so it keeps the conservative edge by design.

### 7. Member/update writes and globalThis aliasing

`obj.x = 1` recorded only a read of `obj`; `globalThis.count++` recorded no write; `const g = globalThis; g.tag = "x"` escaped entirely. Now: member writes through bindings (including destructuring targets like `[obj.x] = arr` and paren/opt-chain forms) bail the statement; update-exprs on global props record write+read cells; a bare global-object alias used as a value taints the bindings it flows into (transitive fixpoint) and statements reading suspects bail.

### 8. window/self

Unshadowed `window`/`self`/`frames`/`top` now share `globalThis`'s cell namespace and bail shapes (README claimed this; the code only matched the literal `globalThis`). Chunk-top-level declarations or imports of those names disable the global treatment. Indirect eval needed no `strip_parens` semantics change — the opaque-call rule subsumes it for the S-chain, and a local `callee_base_expr` helper (parens + comma sequences) feeds the structural eval/Function check.

**Vendor strip interaction:** the strip's swap-privacy gate consumed `dataflow_summarizable` with narrower "which cells does this *write*" semantics; the opaque-call bail would have rejected its pinned droppable-island fixtures. The bit is split: structural `cell_writes_summarizable` (old semantics + indirect eval, for the strip) vs. `dataflow_summarizable` (S-chain, with items 6–8). The strip's global-prop-write-via-alias blind spot is pre-existing and out of scope here.

Tests: `e2e/s_chain_dataflow_soundness_test.rs` (8 fixtures: WAR, console pairs, body global writes, member writes, update-exprs, alias reads, window, indirect eval — each with an owner-graph `Sequenced`-edge assertion plus a runtime output check).

## Docs

README's "Conditionally-correct optimizations" precondition list rewritten to exactly match the implemented checks; design.md's emission-modes section synced (WAR rule, bail list, global aliases) and the at-init promotion section rewritten (resolvable-callee rule, fallback, honest residual limitations).

## Not done (follow-ups)

- Duplicate top-level declarations still last-insert-win in `build_owner_graph_with`'s `binding_owner` map (graph.rs).
- `OwnerGraph::from_report` still silently `continue`s on edges whose endpoints don't resolve.

## Verification

- `bbr test //devinfra/js/debundle/...` — 86/86 green (`e1bcd3b9-53d5-45bb-96aa-94be9792b008`).
- `bbr build //props/frontend/...` — green (no in-repo `debundle_pipeline` consumer exists on this branch; the props/frontend debundle corpus described in AGENTS.md is not present here, so the private-corpus smoke remains for the maintainer).
- Red-first verification: `21e63e08-2390-41e3-8cc3-a314aec3c25d` (all new fixtures failing pre-fix, for the documented reasons).

---

### DIFF
--- a/devinfra/js/debundle/BUILD.bazel
+++ b/devinfra/js/debundle/BUILD.bazel
@@ -249,6 +249,7 @@ rust_library(
         "@crates//:swc_atoms",
         "@crates//:swc_ecma_ast",
         "@crates//:swc_ecma_utils",
+        "@crates//:swc_ecma_visit",
     ],
 )
 
--- a/devinfra/js/debundle/README.md
+++ b/devinfra/js/debundle/README.md
@@ -180,23 +180,45 @@ optimizations" for the soundness rule.
 The first such pass is the dataflow-aware S-chain in `graph.rs`, opted
 into per chunk via
 `chunk_analysis_options.<chunk_id>.dataflow_aware_s_chain` in the spec.
-It is sound when no top-level statement contains:
+Each impure top-level statement carries a `dataflow_summarizable` bit
+(`facts/wire.rs`); a statement keeps the relaxed per-cell emission only
+when it contains **none** of the following (all checked per statement,
+at-init positions only):
 
-- direct `eval(...)` / `(0, eval)(...)`
+- a call, `new`, optional call (`f?.()`), or tagged template the purity
+  classifier does not prove `Pure` — this covers `console.log(...)`
+  and other I/O, calls into chunk functions whose bodies write global
+  props, direct `eval(...)`, **and** indirect `(0, eval)(...)`
+- a member write through a binding: `obj.x = 1`, `obj.x++`,
+  destructuring assignment targets containing member expressions
+  (`[obj.x] = arr`)
 - `with (obj) { ... }`
 - `new Function(...)` / `Function(...)`
-- computed-key `globalThis[<expr>]` / `window[<expr>]` / `self[<expr>]`
-- `Object.defineProperty` / `Reflect.defineProperty` on a global
+- a computed-key access on an unshadowed global-object alias:
+  `globalThis[<expr>]` / `window[<expr>]` / `self[<expr>]` /
+  `frames[<expr>]` / `top[<expr>]`
+- `Object.defineProperty` / `Reflect.defineProperty` with a
+  global-object alias as the target
 - `new Proxy(<global>, ...)`
-- dynamic member-key reads/writes on outer-scope bindings the pass
-  would otherwise track
+- a read of a binding the global object may have escaped into:
+  `const g = globalThis; ... g.tag` taints `g` (transitively, through
+  bindings derived from it), and every statement reading a tainted
+  binding falls back
 
-Each impure top-level statement carries a `dataflow_summarizable` bit
-(`facts/wire.rs`). Statements that fail the check fall back to a
-strict S-edge against every prior impure owner and act as an opaque
-barrier for later statements, so the optimization is safe to enable
-even on bundles that mix audited and unaudited code — only the
-unsummarizable statements pay the conservative cost.
+Statements that fail the check fall back to the strictly-conservative
+S-chain (an edge to every prior impure owner; later statements treat
+them as opaque barriers), so the optimization is safe to enable even
+on bundles that mix audited and unaudited code — only the
+unsummarizable statements pay the conservative cost. Call-heavy
+top-level code therefore sees little benefit from the relaxation:
+every statement containing an unproven call is conservatively
+chained. That is the intended trade — the relaxation only fires where
+the analyzer can actually prove non-interference.
+
+Unshadowed `window` / `self` / `frames` / `top` are treated as the
+same object as `globalThis` for cell tracking: `window.tag = 1` and
+`globalThis.tag` are the same cell. A chunk-top-level declaration or
+import of one of these names disables its global treatment chunk-wide.
 
 See `docs/design.md` → "Emission modes" for the precise dataflow-aware
-emission rule.
+emission rule (including the write-after-read edges).
--- a/devinfra/js/debundle/analysis_tests/mod.rs
+++ b/devinfra/js/debundle/analysis_tests/mod.rs
@@ -691,9 +691,12 @@ sideEffect(C, E);
         assert_eq!(af.first_order_body_calls, bf.first_order_body_calls);
         assert_eq!(af.effects.reads, bf.effects.reads);
         assert_eq!(af.effects.writes, bf.effects.writes);
-        assert_eq!(
-            af.effects.dataflow_summarizable,
-            bf.effects.dataflow_summarizable
-        );
+        // `effects.dataflow_summarizable` is deliberately NOT
+        // compared: the opaque-at-init-call bail consults the purity
+        // classifier (declared-pure hints exempt a call from the
+        // bail), so the bit is policy-DEPENDENT by design. The
+        // structural bail shapes (`with`, dynamic global keys,
+        // member writes, global-object escape taint) remain
+        // policy-independent.
     }
 }
--- a/devinfra/js/debundle/artifact.rs
+++ b/devinfra/js/debundle/artifact.rs
@@ -1671,6 +1671,7 @@ fn dep_kind_key(kind: DepKind) -> &'static str {
         DepKind::LazyUse => "lazy_use",
         DepKind::EagerRebind => "eager_rebind",
         DepKind::LazyRebind => "lazy_rebind",
+        DepKind::DeferredRebind => "deferred_rebind",
         DepKind::Sequenced => "sequenced",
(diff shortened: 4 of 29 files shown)


agentydragon:
Rebased onto `origin/devel` (f3e561f91, includes #2059/#2057/#2058/#2061/#2062) and force-pushed as 356743ee1.

**Conflicts** (only two textual; everything else auto-merged cleanly):

- `README.md` — "Conditionally-correct optimizations" S-chain paragraph. Devel had a shorter version of the fallback paragraph; kept this PR's rewrite (strict fallback + call-heavy-code trade-off + window/self/frames/top aliasing note), dropping devel's duplicate of the `dataflow_summarizable` bit sentence already covered above the conflict.
- `docs/design.md` — "Emission modes" dataflow-aware bullet. Kept this PR's version (adds the write-after-read edge rule), but took devel's corrected module path: `StatementEffectSummary` lives in `facts/mod.rs`, not `facts.rs`.

**Semantic merge verification** (auto-merged hunks checked by hand):

- `compute_shadowed_globals` now combines both changes: #2059's derived `SHADOW_TRACKED_GLOBALS` name-set drives `try_shadow`, fed by this PR's hoisting-aware `collect_declared_names` (`ModuleItem::Stmt → hoisted_var_ids`), so block-hoisted `var Math = shim` registers as shadowing under the derived set.
- Export-default purity routing composes with #2059's stricter classifier signatures (`ExportDefaultExpr → classify_expr_purity`, `ExportDefaultDecl(Class) → class_has_static_observable`, both with the new `local_shadowed`/`declared_pure`/`graph` args). No fixture expectation changes were needed.
- `DeferredRebind`/`G_atomic` and the at-init promotion fallback kept wholesale.

**Tests**: full `bbr test //devinfra/js/debundle/...` green — 95/95, including this PR's red-first fixtures (`at_init_unresolved_callee_test`, `export_default_purity_test`, `hoisted_var_declaration_test`, `s_chain_dataflow_soundness_test`, `at_init_promotion_nested_closure_test`, `at_init_s_chain_dataflow_test`) and the merged PRs' key tests (`purity_test`, `owner_graph_purity_reason_test`, `naturalize_cross_scope_params_test`, `modules_gate_cli_test`, `vendor_swap_test`). Invocation: `buildbuddy:a49db7e5-e242-4d71-a53c-73a6fca13599` (also updated in the commit's `BAZEL_TEST_INVOCATIONS`).

---

PR #2067, opened 2026-06-11
debundle: enforce input-chunk admission checks for A1/A3/A4 assumptions

agentydragon:
## What

`docs/design.md` §"Conditions on the input chunk" calls A1–A5 "statically checkable on each chunk", but only A2 (top-level await) was actually enforced (the TLA bail in `stage_one`). This PR adds an input-chunk admission scan — a new module `chunk_admission.rs` with its own SWC visitor — run by `stage_one::compute_stage_one_analysis` right next to the existing A2 bail, before any quotient or lowering work. Each rejection mirrors the A2 UX: chunk id + offending statement ordinal (post-comma-split view, aligned with owner-graph reports) + the matched shape, plus the override spelling to use after auditing.

## Per-check strength

- **A1 (`eval`) — enforced, module-top only**: direct `eval(...)` and seq-indirect `(0, eval)(...)`; the callee is unwrapped through parens AND comma sequences (no seq-aware helper existed on devel, so the scan carries its own ~10-line unwrap; #2064's, if it lands one, can be unified later). Module code is strict, so an `eval` ident can only be the global eval. **Residual gaps (documented in design.md "Coverage gaps")**: function-body eval — A1's wording only bans module-top eval, but a lazy direct eval can still read cross-module bindings dynamically when called after the split — and aliased eval (`const e = eval`).
- **A3 (internal dynamic import) — enforced**: string-literal `import(...)` specifiers are resolved through `ArtifactIndexes::resolve_runtime_import_reference` (the same resolution the specifier rewriter uses, threaded into Stage A as a callback so Stage A stays artifact-free). A specifier resolving back **into the same chunk** is rejected at any depth — `I` is the per-chunk graph, so that is the "debundled internal module" case. Specifiers resolving to *other* artifact chunks are allowed: design.md explicitly treats vendor/route chunks as black-box leaves, and flagging them would reject every code-splitting bundle (esbuild/SvelteKit `import()` thunks are everywhere) — the worse failure mode. **Non-literal specifiers at top level: bail** (soundness over completeness; no e2e fixture or in-repo corpus relies on them — `props/frontend/debundle/` does not exist in-tree). Non-literal specifiers in lazy positions stay a documented gap; "literal" means `Lit::Str`, matching `prepare_js_chunks` / `rewrite_chunk_entry_specifiers`.
- **A4 (`with`) — enforced at parse, no admission code**: verified experimentally that even with the production parser's `no_early_errors: true` TS syntax, `with` surfaces as a recoverable parse error and `js_ast` fails chunk loading. Per the no-dead-code instruction the visitor check was removed; instead the parse rejection is pinned by `with_block_is_rejected_at_parse`, and `js_ast::parse_module_from_source_file` now includes the recovered errors' messages (previously an opaque "N recoverable parser error(s)"). There is consequently no `a4_with` override — a parse error can't be meaningfully overridden.
- **A5 (namespace reflection) — minimal by design**: only the cheap, low-false-positive sub-shape is checked: module-top `import.meta` use beyond `import.meta.url` (named props like `import.meta.env`, computed `import.meta[...]`, bare `import.meta` as a value). `Function.prototype.toString` reads and `Reflect` descriptor inspection are deliberately **not** checked — a precise analysis would over-flag mainstream bundles. The decision and the unchecked sub-shapes are documented in design.md (A5 bullet + "Coverage gaps").

## Escape hatch

On by default for `debundle run`. Per-chunk, per-check override following the existing `chunk_analysis_options` pattern:

```yaml
chunk_analysis_options:
  static/app:
    admission_overrides: [a1_eval]
```

Every override use prints a one-line notice; an override that no longer suppresses anything warns as redundant (mirroring redundant purity hints). One existing fixture (`at_init_s_chain_dataflow_test::bail_out_keeps_s_edge_when_statement_uses_direct_eval`) intentionally uses top-level eval to pin the dataflow bail-out path — it now opts out via `a1_eval` with a comment.

## Tests

- `e2e/chunk_admission_test.rs`: violating chunk → rejection with the right diagnostic (A1 direct + seq-indirect, A3 same-chunk + top-level non-literal, A4 parse, A5 `import.meta.env`); clean/benign chunk → passes (function-body eval, external + lazy dynamic imports, `import.meta.url`); override → passes with notice; stale override → redundant warning.
- Unit tests in `chunk_admission.rs` for visitor corner cases (paren-wrapped nested seq eval, static-block-eager vs method-body-lazy class positions, template-literal specifiers, ordinal alignment with the comma-split view).
- A2's existing bail remains pinned by `realizability_test.rs::top_level_await_is_rejected` (still present post-merges).
- `spec.rs` tests: override list parsing, unknown check name rejected, empty overrides skipped on serialize.

`bbr test //devinfra/js/debundle/...`: 92/92 green (buildbuddy:98a01fb6-6f75-42a9-a88e-e6805eb98073); `bbr build //...` green.

## Coordination with #2064

Implemented as a new module with its own visitor (no `StatementFactsCollector` changes) and wired in `stage_one/mod.rs` next to the A2 bail, per the conflict-minimization plan — #2064's at-init fallback / hoisted-vars / dataflow-collector work doesn't touch these seams, but **this PR may need a trivial rebase if #2064 merges first** (shared neighborhoods: `lowering/materialize/mod.rs` around the `compute_stage_one_analysis` call, `at_init_s_chain_dataflow_test.rs`). The A3/A1 admission bails are about the strict path's global input assumptions and are complementary to #2064's opaque-call bails in dataflow mode.

---

### DIFF
--- a/devinfra/js/debundle/BUILD.bazel
+++ b/devinfra/js/debundle/BUILD.bazel
@@ -259,6 +259,7 @@ rust_library(
         "analysis_hints.rs",
         "analysis_tests/mod.rs",
         "atomic_units.rs",
+        "chunk_admission.rs",
         "chunk_analysis.rs",
         "chunk_factorization.rs",
         "factor_assembly.rs",
--- a/devinfra/js/debundle/README.md
+++ b/devinfra/js/debundle/README.md
@@ -222,3 +222,35 @@ import of one of these names disables its global treatment chunk-wide.
 
 See `docs/design.md` → "Emission modes" for the precise dataflow-aware
 emission rule (including the write-after-read edges).
+
+## Input-chunk admission checks
+
+Every materialized chunk is screened against the statically checkable
+input assumptions of `docs/design.md` → "Conditions on the input
+chunk" before any quotient or lowering work (`chunk_admission.rs`,
+run from `stage_one::compute_stage_one_analysis` next to the A2
+top-level-await bail). The enforced shapes:
+
+- **A1** — direct `eval(...)` / seq-indirect `(0, eval)(...)` calls at
+  module top level.
+- **A3** — string-literal dynamic `import(...)` resolving back into
+  the same chunk (any depth), and non-literal specifiers at module
+  top level.
+- **A5** (minimal) — `import.meta` use at module top level beyond
+  `import.meta.url`.
+
+A2 (top-level `await`) bails in the same place; A4 (`with`) is
+rejected at parse time. Rejections name the chunk, the offending
+statement ordinal, and the matched shape. For audited corpora,
+disable individual checks per chunk in the spec:
+
+```yaml
+chunk_analysis_options:
+  static/app:
+    admission_overrides: [a1_eval]
+```
+
+Each override prints a one-line notice per run; an override that no
+longer suppresses any violation is reported as redundant (remove it).
+The deliberately unchecked residual shapes are listed in
+`docs/design.md` → "Coverage gaps".
--- a/devinfra/js/debundle/binding_targets.rs
+++ b/devinfra/js/debundle/binding_targets.rs
@@ -256,3 +256,14 @@ pub fn strip_parens(expr: &Expr) -> &Expr {
     }
     cur
 }
+
+/// Look through parens and comma sequences to a callee's final
+/// operand: `(0, eval)(...)` is an indirect eval call with the same
+/// arbitrary-cell-write power as a direct one.
+pub fn callee_base_expr(expr: &Expr) -> &Expr {
+    let mut cur = strip_parens(expr);
+    while let Expr::Seq(seq) = cur {
+        cur = strip_parens(seq.exprs.last().expect("SeqExpr is non-empty"));
+    }
+    cur
+}
--- a/devinfra/js/debundle/chunk_admission.rs
+++ b/devinfra/js/debundle/chunk_admission.rs
@@ -1,0 +1,542 @@
+//! Input-chunk admission scan for the statically-checkable
+//! assumptions of docs/design.md §"Conditions on the input chunk"
+//! (A1 `eval`, A3 internal dynamic `import()`, A5 cheap
+//! `import.meta` reflection shapes). Runs in Stage A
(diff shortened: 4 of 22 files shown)


agentydragon:
Rebased onto current `devel` (`70d0f39a4`, includes #2064) and force-pushed.

**Conflict resolution:**

- `README.md` (only textual conflict): #2064's rewrite of the dataflow-aware emission paragraph ("emission rule (including the write-after-read edges)") and this PR's new "Input-chunk admission checks" section are independent — kept both.
- `docs/design.md`, `stage_one/mod.rs`, etc. auto-merged cleanly; verified the admission scan still runs in `compute_stage_one_analysis` immediately after the A2 TLA bail, and the merged A1–A7 paragraph reads coherently with both PRs' edits.

**Deduplication:** #2064 added a private seq-aware callee unwrap (`callee_base_expr`) in `facts/mod.rs`; this PR carried its own copy in `chunk_admission.rs`. Moved the helper to `binding_targets.rs` (next to `strip_parens`, which both modules already import) as `pub fn callee_base_expr`, and both `facts/mod.rs` and `chunk_admission.rs` now use it. The direct-vs-indirect diagnostic wording is preserved via a `matches!(strip_parens(callee), Expr::Seq(_))` check.

**Semantic collision found by tests:** #2064's new `s_chain_dataflow_soundness_test::indirect_eval_is_a_barrier` fixture uses top-level `(0, eval)(...)`, which this PR's A1 admission check now (correctly) rejects. Added `admission_overrides: [a1_eval]` to that fixture — same pattern as `bail_out_keeps_s_edge_when_statement_uses_direct_eval` — so the per-statement dataflow bail-out stays exercised on its own.

**Tests:** `chunk_admission_test` 11/11, A2 pin (`composer_bails_on_top_level_await`), and #2064's dataflow fixtures (`s_chain_dataflow_soundness_test`, `at_init_s_chain_dataflow_test`, `at_init_unresolved_callee_test`, …) all pass; full `bbr test //devinfra/js/debundle/...` green (96 targets) — invocation `58694079-e89b-43e3-ba95-904d091dc407`.

---

PR #2114, opened 2026-06-11
debundle: Track D — type-shape and strict-mapping cleanups

agentydragon:
Track D of `plans/sanitization_program_2026_06.md` (one PR).

## Landed

1. **`StatementFacts` → `PositionBucketed<T>`** (`facts/mod.rs`, `facts/wire.rs`): the eager/lazy/first-order triples for reads, rebinds, and calls collapse from 9 fields to 3 `PositionBucketed<BTreeSet<Id>>` fields; the `first_order_lazy ⊆ lazy` subset invariant is enforced in a single `record()` helper instead of per-construction-site discipline. `StructuralStatementFacts` now uses the same vocabulary (the field-by-field mid-pipeline renames `at_init_reads`→`eager_reads`, `lazy_calls`→`body_calls`, … are gone). The `effects` cell summary is now derived on demand via `StatementFacts::effects()` — only the `GlobalProp` half (`global_writes`/`global_reads`) and the summarizability bits are stored; the `Binding`-cell half no longer restates `declared`/`reads.eager`/`rebinds.eager`. The now-unused `EffectCellReport`/`StatementEffectSummaryReport` wire mirrors are deleted. Zero behavior change.
2. **`ChunkManifest` → `#[serde(flatten)]` over `ChunkAnalysisReport`** (`artifact.rs`): removes the field-by-field copy. Zero wire change — pinned by a new top-level-key assertion on `chunk.json` in `e2e/realizability_test.rs`. The `ARCHITECTURE_BACKLOG.md` auto-derive entry is resolved and deleted.
3. **`ChunkAnalysis` field privatization** (`chunk_analysis.rs`): all fields private behind accessors so mutating an input can't silently stale the precomputed lookup tables; `facts` and `chunk_renames` are no longer retained past construction. (Started by the previous worker; verified and finished.)
4. **`graph.rs` strict mapping** (failing tests first):
   - `build_owner_graph{,_with}` returns `Err(DuplicateTopLevelDeclaration)` when two top-level statements declare the same binding, instead of last-insert-wins silently dropping every edge into the earlier owner. Surfaced as a spec-facing rejection through stage one; pinned by the new `e2e/duplicate_top_level_declaration_test.rs` (plus a hygiene-negative case: same name in an inner scope is not a duplicate).
   - `OwnerGraph::from_report` returns `Err(UnresolvedOwnerEdgeEndpoint)` on edges whose endpoints don't resolve in the node table, instead of silently dropping them and handing the planner-side gate a weaker graph. Propagated through `QuotientGraph::from_report{,_with_partition,_with_partition_extended}` and `build_seed_quotient` (planner CLI) and `cli/edit_gate.rs`.
5. **`cli/mod.rs` slimming**: the `scc`/`cluster` command implementations, arg structs, and text renderers moved to `cli/scc_cluster.rs` — purely mechanical, no behavior change (kept mechanical per coordination with the Track E worker who owns the CLI gate-rejection/output paths).

`CODE_REVIEW.md` entries for all landed items are deleted; the Track D section of the plan carries a status note.

## Skipped (with rationale)

**Fold `program_analysis.rs` into the facts walk** — not a mechanical fold: `analyze_program_shallow` runs at *prepare* time on a plain (non-resolver) parse of **every** chunk and feeds `artifact::ChunkAnalysisReport` records, while the facts walk runs at *stage one* on resolver-processed ASTs of materialized chunks, in a crate (`analysis`) that doesn't and shouldn't depend on `artifact`. Folding requires redesigning when/where manifest records are computed (or extracting a shared declaration classifier into `binding_targets`). Rationale recorded in the plan's Track D status note; the CODE_REVIEW "Two parallel top-level fact extractors" P1 entry stays open.

## Verification

- `bbr test //devinfra/js/debundle/...` — 101/101 green: `BAZEL_TEST_INVOCATIONS=buildbuddy:11e2cc51-73bc-4327-990d-63f4e3b2dabd`
- The duplicate-declaration e2e was confirmed red against the pre-fix binary (invocation `6c415b36-7982-42c7-81aa-d0bf3856848b`) before the fix landed.
- No `//props/frontend/debundle` targets exist at HEAD (the realistic-shape corpus lives downstream), so the synthetic suite is the in-repo gate.

---

### DIFF
--- a/devinfra/js/debundle/ARCHITECTURE_BACKLOG.md
+++ b/devinfra/js/debundle/ARCHITECTURE_BACKLOG.md
@@ -143,14 +143,6 @@ These files document the same project from multiple perspectives. Skimming them,
 
 ## Concerns to discuss before deciding
 
-### Should `ChunkAnalysisReport` be auto-derived from the IR?
-
-After the rename, `chunk_analysis::ChunkAnalysis` (IR) and `artifact::ChunkAnalysisReport` (JSON wire) still coexist as parallel definitions. The longer-term question is whether the report shape should be derived from the IR shape via a wire-format adapter (the way `OwnerGraph` ↔ `OwnerGraphReport` works). If yes, the two types collapse into one IR + one auto-derived report.
-
-**Decision needed**: whether the report types are auto-derivable from IR types, or whether they intentionally diverge (e.g. the report has fields the IR doesn't, like `parser: ParserOptionsRecord` for reproducibility).
-
-A related collapse with zero wire change is available today: `artifact.rs::ChunkManifest::from_analysis` copies `ChunkAnalysisReport` into `ChunkManifest` field-by-field; embedding the report struct with `#[serde(flatten)]` removes one duplication layer while keeping the serialized shape identical.
-
 ### Gate simulator ↔ materializer import-order sharing (RESOLVED)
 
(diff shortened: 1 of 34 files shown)


agentydragon:
Rebased onto devel over #2112 and re-verified. One conflict: `CODE_REVIEW.md` — both PRs deleted different entries at the same location; resolved as the union of deletions (stringly-chunk-identity entry removed by #2112, `ChunkAnalysis`-privatization entry removed here). Full debundle suite green post-rebase: `BAZEL_TEST_INVOCATIONS=buildbuddy:807d41f8-66c5-478d-ba7a-b5b183e57806`. Committer identity reset per the verification hook.

---

PR #2128, opened 2026-06-11
[codex] loom: resolve wayback captures via availability API

agentydragon:
## Summary

- resolve normal Wayback proxy lookups through `archive.org/wayback/available` before falling back to clamped CDX
- route/cache `/wayback/available` separately from replay/CDX in `wayback-cache`, with origin-aware cache keys and IA signal-header metrics
- update fake IA coverage, proxy tests, and the Loom handoff/API docs for the new proxy/cache split

## Why

The eval was spending a large amount of cold traffic on `web.archive.org/cdx/search/cdx`, where IA was returning many `502`/`504` failures. Availability gives the proxy a cheaper single-capture lookup for the common path while keeping CDX available for passthrough and fallback cases such as archived non-200 captures.

## Validation

- `git diff --check`
- `bbr test //loom/wayback_proxy:test_proxy //loom/wayback_proxy:test_compose_e2e //loom/wayback_proxy:test_live_ia`

Note: I could not run local `nginx -t` because `nginx` is not installed in this environment; the cache config was reviewed statically and covered by the existing repo hooks where applicable.

### DIFF
--- a/cluster/k8s/wayback-cache/config/nginx.conf
+++ b/cluster/k8s/wayback-cache/config/nginx.conf
@@ -3,12 +3,12 @@
 #   tier 1 (:8080)          — PVC-backed proxy_cache; hits never touch IA.
 #   tier 2 (127.0.0.1:8081) — loopback-only; every request reaching it is by
 #                             construction a cold miss, so limit_req buckets pace
-#                             exactly the traffic IA sees — content fetches and
-#                             the slower CDX index queries throttled separately
-#                             so neither starves the other.
-# The cache is deliberately dumb: any path forwards verbatim to
-# web.archive.org (CDX queries and /web/<ts>id_/ fetches alike). Date
-# clamping is the per-agent wayback proxy's job (loom/wayback_proxy/).
+#                             exactly the traffic IA sees — replay content,
+#                             Availability lookups, and slower CDX index queries
+#                             are throttled separately so none starves another.
+# The cache is deliberately policy-dumb: /wayback/available forwards to
+# archive.org; /web/... and /cdx/... forward to web.archive.org. Date clamping is
+# the per-agent wayback proxy's job (loom/wayback_proxy/).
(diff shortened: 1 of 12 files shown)


agentydragon:
Follow-up pushed in `3ac6d9143`:

- Added Pydantic models for the Wayback Availability response shape in `loom/wayback_proxy/proxy.py`.
- Kept the proxy-owned as-of policy check (`timestamp <= WAYBACK_AS_OF`) after parsing, so `closest` captures after the clamp still return `None` and fall back to CDX.
- Added `@pypi//pydantic` to the proxy target and a malformed Availability-shape unit test.
- Merged current `origin/devel` into the PR branch without rewriting the already-pushed commit history.

Validation:

- `bbr test //loom/wayback_proxy:test_proxy` passed.
- `bbr test //loom/wayback_proxy:test_compose_e2e //loom/wayback_proxy:test_live_ia` passed before commit.
- After merging current `devel`, `bbr test //loom/wayback_proxy:test_proxy //loom/wayback_proxy:test_compose_e2e //loom/wayback_proxy:test_live_ia` had unit/compose pass from cache, but the manual live IA target timed out at Bazel's 60s limit.
- Rerun of `bbr test //loom/wayback_proxy:test_live_ia --test_output=streamed --nocache_test_results` passed (`1 passed in 10.32s`).

agentydragon:
CI follow-up pushed in `91e03a002`.

Root cause: `_emit_upstream_error()` accepted `headers: object | None`, but then called `.get()`, so the mypy aspect for `//loom/wayback_proxy:proxy` failed with `[attr-defined]`.

Fix: annotate the concrete aiohttp header type, `multidict.CIMultiDictProxy[str]`, and add the direct `@pypi//multidict` Bazel dep.

Validation:

- `bbr build //loom/wayback_proxy:proxy` passed.
- `bbr test //loom/wayback_proxy:test_proxy //:requirements_test` passed.

PR #2299, opened 2026-06-16
debundle: key-set minimization for objects in var groups (#2290)

### LINKED ISSUE #2290: debundle: object key-set pin minimization
## Problem

When an object is discriminated by its **key set** rather than any value — every value already holed to `ANYTHING` — the minimizer keeps **every** key. Examples from the gaffer dogfood:

```js
const workspaceLoadErrorPageStyles = {
  diagnosticsSection: ANYTHING, detailsToggle: ANYTHING, details: ANYTHING,
  detailRow: ANYTHING, detailLabel: ANYTHING, detailValue: ANYTHING,
  logSection: ANYTHING, logs: ANYTHING, actions: ANYTHING,
  primaryButton: ANYTHING, secondaryButton: ANYTHING
};
```
and a schema object `aiMessageDataSchema = ee({ coreMessage: …, type: …, … })` kept whole. None of the values discriminate, so the object is identified purely by which keys are present.

## Fix

Keep only the **minimal key subset** whose presence still discriminates the target from its siblings, and collapse the rest to `OBJECT_PROPS` — the key-set analogue of greedy set-cover: anchor on the rarest discriminating keys, `OBJECT_PROPS` the common ones. (Today the single-target object read-off keeps a discriminating `key: value`; this is the all-values-holed variant where the key *set* is the discriminator.)

Note these over-pins surface inside **multi-declarator var groups** (`DECLARATORS_BEFORE`/`_AFTER`), so the key-set minimization must run on the per-declarator object even when the statement is a declarator group.

## Tracking

- Plan: `devinfra/js/debundle/plans/readoff_minimization.md`, sub-item under the Object-dict family backlog (item 5).
- Related: interior holing (#2289) — the recursive-holing machinery overlaps; key-set minimization is the "which keys to keep" decision on top of it.
- Impact: contributes to the ~200 convertible-but-bulky var/object pins (the plurality of the cheap-recovery bucket) in the measured debt breakdown.

https://claude.ai/code/session_01RYUAvJefa2eJRgZ7XBfqxA

agentydragon:
Closes #2290.

## Problem

When a target object inside a multi-declarator `var` group is discriminated purely by its **key set** (every value already holed to `ANYTHING`), the group cover added *every* object key at once through the coarse `structural` tier, keeping all keys held to `ANYTHING`.

## Finding

Built the CLI and reproduced it. The **single-declarator** object form *already* minimized key-sets correctly via `minimal_anchor_set` (object keys are scored features, so greedy set-cover already covers them) — confirmed for both a single rare key and a true 2-key cover. The actual gap was only the **multi-declarator var group** path: a target keeping all 4 keys when only `logSection` discriminated it from a same-shape sibling group.

## Fix

Isolated to the group path:

- **`shape_index.rs`** — `selector_feature_selectivity` (chunk-wide posting-list size) to rank keys by rarity.
- **`readoff_render.rs`** — exposed `object_key_label_span` for reuse.
- **`selector_codemod.rs`** — `minimize_var_group_selector` now runs a **greedy key-set set-cover** (`ranked_object_key_anchors_for_slots`) before the structural tier: it adds each target slot's object keys **rarest-first**, re-proving with the matcher oracle, and stops at the minimal discriminating subset. Direct object-literal inits render through `hole_object_padded` so kept keys match as interior `OBJECT_PROPS` subsequence elements and survive key reorder.
- **`plans/readoff_minimization.md`** — marked the sub-item landed.

Result on the reproduced case:

```js
var SelectedStyles = {
    OBJECT_PROPS,
    logSection: ANYTHING,
    OBJECT_PROPS
}, DECLARATORS_AFTER = null;
```

(was: all 4 keys held to `ANYTHING`).

## Tests

- New active E2E `object_keyset_group` (single rarest key in a var group).
- Unit tests in `selector_codemod.rs` for the single-key and true 2-key covers, each asserting unique resolution via the matcher.
- All 116 `//devinfra/js/debundle/...` tests pass; pre-commit (rustfmt / prettier / ducktape hooks) clean.

## Out of scope (distinct, still-open object-dict items, noted in the plan)

- Value-discriminated grouped objects keeping all *value* literals (`grouped_enum_objects` — a value-vs-key anchor-preference question).
- `ee({ … })`-style schema objects passed as a call argument rather than a direct init.

https://claude.ai/code/session_013ikMhehN17gsq2JaAZF2WJ

---
_Generated by [Claude Code](https://claude.ai/code/session_013ikMhehN17gsq2JaAZF2WJ)_

### DIFF
--- a/devinfra/js/debundle/e2e/selector_minimizer_expectation_test.rs
+++ b/devinfra/js/debundle/e2e/selector_minimizer_expectation_test.rs
@@ -377,6 +377,24 @@ minimizer_expectation_case!(
     expected = "expected_match.js",
 );
 
+// A target object inside a multi-declarator var group, discriminated purely by
+// its KEY SET (every value is a non-discriminating call, alpha-wildcarded),
+// minimizes to OBJECT_PROPS holes around the single rarest discriminating key
+// (`logSection`, which the same-shape sibling group lacks) plus DECLARATORS_AFTER
+// for the trailing helper declarator. Previously the group cover added every
+// object key at once (the coarse structural tier), keeping all keys held to
+// ANYTHING; the key-set greedy cover (#2290) adds keys rarest-first and stops at
+// the minimal discriminating subset. Real-spec analogues: CSS-styles dicts and
+// schema objects kept whole inside `var a={…},b={…},…` groups.
+minimizer_expectation_case!(
+    minimizes_object_keyset_group,
+    fixture = "object_keyset_group",
+    name = "key-set-discriminated object in a var group keeps only the rarest discriminating key",
+    module = "app/styles",
+    bindings = [("SelectedStyles", "selectedStyles")],
+    expected = "expected_match.js",
+);
+
 // A subclass among many sibling subclasses of a shared base minimizes to the
 // `extends` clause (superclass holed with ANYTHING) plus the single
 // discriminating class field, with CLASS_REST holes absorbing every other
--- a/devinfra/js/debundle/e2e/testdata/selector_minimizer_expectations/object_keyset_group/expected_match.js
+++ b/devinfra/js/debundle/e2e/testdata/selector_minimizer_expectations/object_keyset_group/expected_match.js
@@ -1,0 +1,6 @@
+var SelectedStyles = {
+    OBJECT_PROPS,
+    logSection: ANYTHING,
+    OBJECT_PROPS,
+  },
+  DECLARATORS_AFTER = null;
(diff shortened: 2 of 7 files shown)


agentydragon:
Closing as redundant: #2290 already landed on `devel` via #2293 (commit 5ea00fa5), a more complete implementation developed in parallel.

#2293 supersedes this PR:
- Its `try_object_read_off` covers single-target objects both standalone and inside multi-declarator groups (slot-aware cover), whereas this PR's group-path greedy uses **edge-only** `OBJECT_PROPS` padding that fails non-adjacent multi-key sets (e.g. keep `alpha`+`delta`, drop `beta`/`gamma`) and falls back to keeping every key. #2293 interleaves `OBJECT_PROPS` between every kept prop, handling those.
- #2293 also unignored `grouped_enum_objects`, which this PR left out of scope.

The only case this PR touches that #2293 doesn't is key-set minimization for *multi-target* binding groups (`targets.len() > 1`) — rare, handled worse here, and already covered conceptually by the plan's multi-target binding-group migration item. Not worth a conflicting parallel implementation.

No rebase/force-push performed.

---

PR #2668, opened 2026-06-29
haku/console: collapse to a bare trusted shell; haku-ui owns the chrome

agentydragon:
Re-splits the console / haku-ui boundary so the **trusted console does only its irreducible job** — frame Haku's UI and broker privileged requests — and all product chrome lives in **haku-ui**. The matching haku-ui side already shipped (it gained the title bar, a global "Note to Haku" FAB, and a launch dialog that posts `requestLaunch`).

## Frontend
- **Full-page iframe.** Drop the console's title bar, its launch button, and the "Note to Haku" feedback FAB — those now live in haku-ui.
- **`requestLaunch` bridge.** The agent-authored iframe can only *ask* to launch: it posts `requestLaunch{prompt}`; the shell renders its **own** top-layer confirm (prompt shown verbatim) and only then fires the capability. Launch authority stays on trusted-rendered chrome — agent UI can request it but never script or spoof it. `ConfirmDialog` gains a `preview` slot for the prompt; `bridge.ts` gains `requestLaunch`/`launchResult` next to `openLink` (parse tests added).
- Remove now-dead `feedback.tsx`, `launch.tsx`, `assets.ts` + their build wiring and the tabler-icons typing; drop `postTrace` from the client.

## Backend — the trace tier is deleted (the console's only git use)
- Remove `trace.py`, `git_state.py` (pygit2 clone/commit/push), `TraceRequest`, the `GitState` dep wiring, and the git_* settings; `create_app` no longer clones haku-state.
- The `@trixie_ca_certificates` image layer **stays** — the capability tier's outbound HTTPS to `api.anthropic.com` needs it (comment corrected; it previously credited pygit2).

## Infra — the console sheds the `haku-state-git-write` credential entirely
- Drop the git env vars + the `/data` clone volume from the Deployment.
- Drop `haku-console` from the `haku-state-git-write` secret's `for_each` in `tf/gitops/haku-state` (kept for `haku-sandbox` + `flux-system`). **Operator note:** this removes the console's copy of the secret on the next tofu apply — please eyeball that change.

## Verification
- Backend tests pass locally via uv (14: config/CSP/cache + the full capability tier); the bridge parse tests cover `requestLaunch`. 
- **The Bazel build + frontend `tsc`/vitest gates could not run in this session** (the env network policy 403s `releases.bazel.build`, so bazel can't download) — relying on CI for those.
- Docs updated to match: `README.md`, the iframe plan's status banner, and the `app.py`/`capabilities.py` module docstrings.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WPYQfvQ9q4NEwNwpjnYR8J)_


haku/console/frontend/app.tsx
```diff
@@ -41,23 +41,14 @@ export default function App() {
         <Loader />
       </div>
     );
+  if (!config.haku_ui_url)
+    return (
+      <Text c="red" className="mx-auto max-w-3xl p-4">
+        Haku UI URL is not configured.
```
agentydragon: let's make it required. it can always be about:blank if unused

haku/console/frontend/bridge.ts
```diff
@@ -8,18 +8,29 @@
...
+  | { type: "requestLaunch"; id: string; prompt: string };
 
 // Outbound result (shell → iframe), so Haku's UI can react to the outcome.
-export type Outbound = { type: "openLinkResult"; url: string; opened: boolean; reason?: string };
+export type Outbound =
+  | { type: "openLinkResult"; url: string; opened: boolean; reason?: string }
+  | { type: "launchResult"; id: string; ok: boolean; sessionUrl?: string; reason?: string };
 
 // Narrow an untrusted postMessage payload to a known message, or null.
 export function parseInbound(data: unknown): Inbound | null {
```
agentydragon, line 25: can we use zod or something like that?

haku/console/frontend/confirm_dialog.tsx
```diff
@@ -8,6 +8,7 @@ export interface ConfirmRequest {
   title: string;
   body?: string; // what the action does, in trusted shell copy
   url?: string; // for openLink — the full URL shown verbatim
+  preview?: string; // verbatim agent-supplied text to review (e.g. a launch prompt)
```
agentydragon: actually the way a given escalation action should be rendered varies by its type i think. i think this should be a disjoint union, or maybe even hold the actual action

haku/console/frontend/bridge.ts
agentydragon: Held off on zod for now, but happy to add it. The blocker is mechanical: zod isn't a repo dep yet, so adding it means a `package.json` + `pnpm-lock.yaml` change, and the lockfile is Bazel-managed (`@aspect_rules_js`) — I can't regenerate it in this session (the env's network policy 403s the bazel binary download, so no Bazel runs here). Adding zod blind would leave `bazel-ci` red on the lockfile until a Bazel-capable run lands it.

So for this PR I kept the validation hand-rolled but made it a proper **discriminated union** (`Inbound = openLink | requestLaunch`), and `parseInbound` narrows exhaustively on `type` — and the confirm dialog now holds that typed action (your other comment). If you'd like zod, two options: (a) I add it now and we let a follow-up / your local run regen the lockfile, or (b) I open a tracked follow-up to swap `parseInbound` to `z.discriminatedUnion` once Bazel's reachable. Which do you prefer?

---

PR #2708, opened 2026-07-01
devinfra: propose bb remote --repo_url patch, build patched bb via bzlmod

agentydragon:
## Summary

- `third_party/buildbuddy_cli/bb_repo_url_commit_override.patch`: a small,
  backward-compatible patch to `bb remote` (`buildbuddy-io/buildbuddy`)
  adding one new, opt-in flag:
  - `--repo_url`: use this URL when fetching on the remote runner instead
    of the local remote's own URL. This is a **pure URL substitution** —
    `determineRemote()`/`determineDefaultBranch()`/`getBaseBranchAndCommit()`/
    `generatePatches()` still run exactly as today, resolving
    branch/commit/patches against the local remote's own tracking refs
    (e.g. `origin`'s); only the final fetch URL sent to the runner is
    swapped. So `bb test`/`bb build` keep working with zero extra flags
    once the override is set — no need to also pass `--run_from_commit`
    just to point at a different URL. Falls back to a new
    `buildbuddy.remote-bazel-repo-url` `.git/config` key when the flag
    itself is unset, mirroring how `bb` already caches the picked remote
    name under `buildbuddy.remote-bazel-remote-name` — so session-setup
    tooling can configure the override once instead of passing the flag
    on every invocation.

  Verified against the real `buildbuddy-io/buildbuddy` source (cloned at
  commit `d4e8918`): the patch applies cleanly to a fresh clone, is
  `gofmt`-clean, includes new test cases (`TestGitConfig_RepoURLOverride`)
  in the existing test harness style, and — now that the devshell build
  below exists — actually compiles and runs (`bb help remote` shows the
  new `-repo_url` flag with its full help text).

- `third_party/buildbuddy_cli/`: a self-contained bzlmod submodule
  (`local_path_override`'d from root `MODULE.bazel`) that vendors
  `buildbuddy-io/buildbuddy` as a `go_deps`-resolved Go dependency, so
  `//third_party/buildbuddy_cli:bb` builds a real `bb-linux-amd64` binary
  with the `--repo_url` patch applied — via `bbr build
  '@ducktape_buildbuddy_cli//:bb'` — instead of consuming BuildBuddy's
  prebuilt release binary. Building `cli/cmd/bb` under ducktape's own
  Bazel graph (rather than buildbuddy's own, which we don't import)
  needed a number of devshell-only patches, layered via
  `go_deps.module_override`/`single_version_override`/`gazelle_override`
  in root `MODULE.bazel`:
  - Stripping BUILD-file references to repos unreachable from the fetched
    repo's context: module-extension-generated repos only resolve
    `@repo` labels visible to the *extension's hosting module*
    (`bazel_gazelle`, since `go_deps` is
    `@bazel_gazelle//:extensions.bzl%go_deps`) — not labels from `bazel_dep`
    entries declared by root or the calling submodule. Affected:
    `rules_python`, `rules_shell`, `rules_pkg`, `with_cfg.bzl`, `io_kythe`
    (whose `go.mod` `replace` target isn't honored by `go_deps.from_file`),
    and `vtprotobuf`'s `ext.proto` include path (a pure marshal/pooling
    perf annotation, safe to drop).
  - Dropping the `bb fix` and `bb index` subcommands: both embed
    dependencies (a different gazelle major version, go-tree-sitter's
    cgo-only types) incompatible with a pure, cross-compiled build under
    ducktape's own toolchain versions, and neither is related to
    `bb remote`.
  - Adding `googleapis` as a `bazel_dep` to gazelle's own `MODULE.bazel`
    (via `single_version_override`, since gazelle doesn't declare it) so
    buildbuddy's real REAPI/bytestream/longrunning proto usage resolves,
    instead of stripping that (actually load-bearing) functionality.
  - A `gazelle_override` for `prometheus/client_model` disabling proto
    generation (the checked-in `go/metrics.pb.go` already is that
    `.proto`'s compiled output; letting gazelle also generate a
    `go_proto_library` from the source produced two targets claiming the
    same Go import path) plus redirecting buildbuddy's hand-written
    `io/prometheus/client:go` label references to the resulting `go:go`
    target.
  - `//third_party/buildbuddy_cli` is simultaneously a subpackage of root's
    own file tree and the root package of the separate
    `@ducktape_buildbuddy_cli` module. Excluded it from root's own
    `//...` expansion via `.bazelignore` (matching the existing
    `mcporter`/`activitywatch`/`manifold_mcp_server` convention) since
    root can't resolve `@com_github_buildbuddy_io_buildbuddy` the way the
    submodule's own context can; moved the patch files root's
    `MODULE.bazel` needs into the existing `third_party/patches/`
    directory so they stay resolvable.
  - Also needed the repo's `go_sdk` toolchain bumped to 1.26.4
    (buildbuddy's `go.mod` requires it) — landed separately in #2716
    (merged), which this PR now rebases on top of.

- `devinfra/plans/bbr_direct_buildbuddy_api.md`: a shelved alternative —
  `bbr` bypassing `bb remote` entirely and talking to BuildBuddy's Twirp
  API directly via the `bbapi` tool already in this repo. Recommendation
  in the doc is not to build it now; the patch above covers the actual
  need at much lower cost.

An earlier revision also proposed `--apply_local_patches` (mirror local
changes even when pinning an explicit `--run_from_branch`/
`--run_from_commit`). Dropped: its only justification was pairing with
`--repo_url` back when that flag required `--run_from_commit`. Once
`--repo_url` became a pure substitution with no such requirement,
`--apply_local_patches` had no concrete driving use case — keeping the
upstream ask to just `--repo_url` keeps it minimal and clearly motivated.

## Sequencing

This PR adds the proposal/patch and a working devshell build of the
patched `bb`. Not yet done (follow-up, not here): wire the patched binary
into the actual Nix devshell (`nix/packages/bb.nix` currently fetches a
prebuilt release binary) — likely via a new CI workflow that builds and
publishes it as a ducktape GitHub Release, then repoints
`nix/artifact-pins.json` at it. After that, a separate PR switches
`devinfra/claude/reconcile_bbr_remote.sh` to write
`buildbuddy.remote-bazel-repo-url` on the existing `origin` remote instead
of maintaining the separate `github-no-proxy` remote.

## Test plan

- [x] Patch applies cleanly to a fresh `buildbuddy-io/buildbuddy` clone
      (`git apply --check`)
- [x] `gofmt -l` clean on both modified files
- [x] New `TestGitConfig_RepoURLOverride` test cases follow the existing
      harness style
- [x] `bbr build '@ducktape_buildbuddy_cli//:bb'` succeeds end to end
      (clean/`--keep_going` and normal cached runs)
- [x] Resulting binary runs (`bb version`) and exposes the new flag
      (`bb help remote` shows `-repo_url` with its full help text)
- [x] `bazel test //...` and `bazel build //...` (CI's exact flags) pass
      locally
- [ ] Not yet submitted upstream

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jp4qPMQ8AEHWzXoRVZMMc7)_


devinfra/plans/bb_repo_url_commit_override.md
```diff
@@ -0,0 +1,132 @@
+# Proposal: `--repo_url` / `--apply_local_patches` flags for `bb remote`
```
agentydragon: Use underscores in filenames preferably. Not dashes

PR #2754, opened 2026-07-02
haku: dispatcher + zone-worker image (dispatch step 4)

agentydragon:
Build-order step 4 of `haku/plans/multi_agent.md`, on top of the merged zone/dispatch infra (#2748). Squashed base commit plus the classifier-via-LiteLLM follow-up; naming follows the operator-directed terminology (dispatcher, prompts, zones — no tier codes).

## The dispatcher (`haku/dispatch/`)

FastAPI service patterned on `airlock/` — the one bespoke component in the dispatch plane:

- **`POST /jobs`** (Haku bearer): deterministic credential-material lint → Anthropic classifier (tool-forced `ClassifierVerdict`; the system prompt is the zai-zone admission policy, reviewed like code) → mint per-job virtual key on workers-LiteLLM (zone model allowlist + budget + TTL) → stamp Job + same-named Secret from the reviewed template. Idempotency: Job name = `job-<sha256(idempotency_key)[:16]>`; the API server arbitrates races — 409 on Job create and `IntegrityError` on the row insert both resolve to the winner. Rejections return a 403 `RejectionResponse` via an exception handler, so the OpenAPI schema matches the wire.
- **`POST /jobs/<id>/result`**: worker turn-in authenticated by a deterministic HMAC job token (`HMAC(secret, job_id)` — no token storage).
- **`DELETE /jobs/<id>`**: kill switch — delete Job/Secret, revoke the LiteLLM key.
- **No GET surface** (operator, 2026-07-02): Haku reads the `jobs`/`results` tables via the `haku_reader` read-only role — a CNPG managed role in `pg_read_all_data`, fully declarative; credential (ESO-generated, plaid-mcp pattern) reflects into `haku-sandbox`. Cross-DB caveat documented: the role can read the `litellm_workers` DB too, which holds only hashed virtual keys and spend logs.
- **Classifier goes through the main LiteLLM** (operator direction): `claude-{sonnet-5,haiku-4-5}` registered on the main instance (upstream key = ESO mirror of the spend-capped `haku-cloud-anthropic-api-key`, now living only in the `litellm` namespace); the dispatcher holds a claude-\*-only virtual key (`tf/gitops/litellm-keys`, reflected into `haku-dispatch`) and points the anthropic SDK at LiteLLM via `ANTHROPIC_BASE_URL`. Langfuse logging, budget, and kill switch for free — no raw provider credential in `haku-dispatch`, and the dispatcher CNP has no external FQDN egress at all.

Jobs carry a self-contained **prompt** (no `workspace_repo` field — a job that needs a repo is told in the prompt to clone it).

Tests: HMAC roundtrip, lint patterns, **template-stamping parity against the real `job-template.yaml`** (placeholders all filled, secrets only via `secretKeyRef`, `automountServiceAccountToken: false`), and sqlite-backed endpoint tests with fake stamper/keys/classifier.

## Worker image (`haku/dispatch/worker/`)

`node:22-bookworm-slim` + Claude Code CLI `2.1.198` + Codex CLI `0.142.5` + git + python3, with a **stdlib-only** entrypoint: run the zone harness headless on the prompt (cwd inside the clone), then POST `/output/result.md` + exit status. Built/pushed as `ghcr.io/agentydragon/haku-zone-worker` by its own workflow (npm installs ⇒ Dockerfile, not `oci_image`). The image holds no credentials.

## Cluster wiring

- `haku/dispatch/dispatcher/`: Deployment (secrets via `secretKeyRef` only — pod specs are Haku-visible), Service, SA + Role/RoleBinding **in** `haku-sandbox-zai` scoped to jobs+secrets, CNP (ingress `haku-sandbox` + zone pods on :8000; egress DNS/workers-LiteLLM/DB/kube-apiserver/main LiteLLM), ESO-generated Haku token + result-HMAC, Job template via `configMapGenerator`.
- `haku/dispatch/db/`: `dispatcher` database as a CNPG `Database` CR (first in repo — chosen over app-owned `CREATE DATABASE` to stay declarative) + `managed.roles` (`dispatcher` owner, `haku_reader`) with ESO `Password`-generator basic-auth secrets.
- `cluster/k8s/litellm/`: anthropic models in the generated `proxy-config.yaml` (parity-tested), `ANTHROPIC_API_KEY` env + `secrets/anthropic-api-key-eso.yaml`.
- `haku/zones/zai/`: no-grants `haku-zone-worker` ServiceAccount.
- Flux image automation (ImageRepository/ImagePolicy + webhook receiver) for `haku-dispatcher` and `haku-zone-worker`; `push-images.yml` matrix row for `haku-dispatcher`.

## Notes / residue

- Classifier default model `claude-sonnet-5`; the shared spend cap with L0's workspace is deliberate v1 (a dedicated upstream key needs an operator console trip).
- First-prompt end-to-end smoke (two-hop `/v1/messages` streaming through workers-LiteLLM) happens at rollout per the plan.

https://claude.ai/code/session_01BDFQogCNq94uP7rdiYYBZg


haku/dispatch/db.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class JobRow(Base):
+    __tablename__ = "jobs"
+
+    id: Mapped[str] = mapped_column(String(63), primary_key=True)
+    zone: Mapped[str] = mapped_column(String(16))
+    model: Mapped[str] = mapped_column(String(100))
+    brief: Mapped[str] = mapped_column(Text)
+    workspace_repo: Mapped[str | None] = mapped_column(String(500), nullable=True)
```
agentydragon: i don't quite see why we should have a workspace_repo. if we want the agent to use git, we can just tell it to clone git.

haku/dispatch/db.py
```diff
@@ -0,0 +1,119 @@
...
+    status: Mapped[str] = mapped_column(String(16), index=True)
```
agentydragon: enum?

haku/dispatch/db.py
```diff
@@ -0,0 +1,119 @@
...
+    pass
+
+
+class JobRow(Base):
+    __tablename__ = "jobs"
+
+    id: Mapped[str] = mapped_column(String(63), primary_key=True)
+    zone: Mapped[str] = mapped_column(String(16))
+    model: Mapped[str] = mapped_column(String(100))
+    brief: Mapped[str] = mapped_column(Text)
```
agentydragon: i think 'task' or 'prompt' might be more obvious

haku/dispatch/db.py
```diff
@@ -0,0 +1,119 @@
...
+async def grant_l0_read(engine: AsyncEngine, role: str) -> None:
+    """SELECT-only access for L0's read-only role (operator, 2026-07-02: L0
+    reads jobs/results via SQL instead of a GET API). The role itself is a CNPG
+    managed role on haku-dispatch-db; only the grants live here, next to the
+    schema they cover. Postgres-only — tests run sqlite and skip this."""
+    async with engine.begin() as conn:
+        for statement in (
+            f'GRANT USAGE ON SCHEMA public TO "{role}"',
+            f'GRANT SELECT ON ALL TABLES IN SCHEMA public TO "{role}"',
+            f'ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO "{role}"',
```
agentydragon: i think this should not even live in python, i'm pretty sure there's a way to provision read-only users in cnpg manifests

haku/dispatch/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class ResultSubmission(BaseModel):
+    result: str = Field(description="Contents of /output/result.md (empty if the agent produced none).")
+    exit_code: int
+
+
+class ClassifierVerdict(BaseModel):
+    allowed: bool
+    reason: str = Field(description="Shown to L0 verbatim on rejection so it can revise the brief.")
```
agentydragon, line 53: i'd probably not use L0/L* naming in the implemnetation.

haku/dispatch/db.py
agentydragon: Done — the grants left Python entirely. `haku_reader` is now a CNPG managed role with `inRoles: [pg_read_all_data]` (CNPG's only declarative read-grant mechanism; per-table GRANTs aren't expressible in manifests). One caveat, documented in `postgres-cluster.yaml`: `pg_read_all_data` spans every database in the cluster the role can connect to, so `haku_reader` can also read `litellm_workers` — which holds only sha256-hashed virtual keys and spend logs, never plaintext credentials (models/creds stay out of the DB via `store_model_in_db: false`).

---

haku/dispatch/config.py
```diff
@@ -0,0 +1,56 @@
...
+
+# Zone wiring. Model allowlists must match the zone key minted in
+# tf/gitops/litellm-keys/main.tf and the model_name entries in
+# cluster/k8s/haku/dispatch/litellm/workers-litellm-config.yaml.
+ZONE_NAMESPACES: dict[Zone, str] = {Zone.ZAI: "haku-sandbox-zai"}
+ZONE_MODELS: dict[Zone, set[str]] = {
+    Zone.ZAI: {
+        f"{m}-anthropic"
+        for m in ["glm-4.5", "glm-4.5-air", "glm-4.6", "glm-4.7", "glm-5", "glm-5-turbo", "glm-5.1", "glm-5.2"]
+    }
```
agentydragon: This should be I think configurable

haku/dispatch/worker/entrypoint.py
```diff
@@ -0,0 +1,74 @@
...
+    prompt = Path(os.environ.get("PROMPT_PATH", "/prompt/prompt.md")).read_text()
+
+    _WORKSPACE.mkdir(exist_ok=True)
+    _OUTPUT.parent.mkdir(exist_ok=True)
+
+    exit_code = _run_harness(os.environ["HARNESS"], prompt, os.environ["MODEL"])
+
+    result = _OUTPUT.read_text() if _OUTPUT.exists() else ""
+    if not result:
+        logger.warning("agent wrote no /output/result.md; submitting empty result")
```
agentydragon: might be useful to instead send some descriptive text like 'agent didn't write /output/result.md'

haku/dispatch/classifier.py
```diff
@@ -0,0 +1,71 @@
...
+agent with access to its operator's private life) to worker agents running on an
+UNTRUSTED external LLM provider. Everything in the prompt will be sent to that
+provider; treat the provider as an adversary that logs and mines all input.
+
+The zone under review is "{zone}". Zone "zai" (z.ai, PRC jurisdiction) admits
+only public-by-construction content.
+
+REJECT the prompt if it contains ANY of:
+- Names, usernames, email addresses, physical addresses, phone numbers, or
+  employer of the operator or anyone in their life (public figures in public
```
agentydragon: not sure how classifier would know what my name is / who is in my life

haku/dispatch/classifier.py
```diff
@@ -0,0 +1,71 @@
...
+  context are fine).
+- Health, financial, legal, or relationship information about a person.
+- Schedules, locations, travel plans, or daily routines of a person.
+- Credentials, tokens, internal hostnames, or non-public URLs.
+- Contents of, or references that reveal contents of, private repositories,
```
agentydragon: also, not sure how classifier would know what code is in private/public repos, or what's in my notes/documents (so it knows what to accept/reject)

wonder if maybe the prompt or part of it should be configurable

haku/dispatch/config.py
agentydragon: Done — zone wiring (namespace + model allowlist per zone) moved out of code into `zones.yaml`, configMapGenerator-mounted next to the Job template. The `Zone` enum became a plain string validated against the loaded config (unknown zone → 422). New `test_zones_config` pins the deployed `zones.yaml` to the workers-LiteLLM generator's model list, closing the parity loop with the existing TF-allowlist check.

---

haku/dispatch/worker/entrypoint.py
agentydragon: Done — missing `/output/result.md` now submits `(agent did not write /output/result.md; see pod log for harness output)` instead of an empty string.

---

haku/dispatch/classifier.py
agentydragon: Fair — it can't. Rewrote the base policy to judge by **shape**, not identity: personal-looking data about *any* private individual is rejectable (names in personal context, contact details, health/finance/schedule facts), so the classifier never needs to know who you are. For the cases where private specifics *would* sharpen it (names that must never appear, etc.), there's now an optional operator-context file — see the sibling thread.

---

haku/dispatch/classifier.py
agentydragon: Same treatment: the base policy now rejects on **provenance** — pasted code/text without a public source reference, or content that reads like notes/email/chats/tickets, is rejected because the classifier can't verify it's public; content referenced by public URL is accepted. And yes, part of the prompt is now configurable: the deployment mounts an **optional** Secret `dispatcher-classifier-context` (key `context.md`) whose text is appended to the system prompt as operator-provided context — e.g. identifiers that must never appear, or repos known to be public. Absent secret = base policy only; when you want it, author the Secret with `sops -e -i` in the dispatcher dir (private content never lands in this public repo). Path is also overridable via `CLASSIFIER_CONTEXT_PATH`.

---

PR #2780, opened 2026-07-04
revert(haku): self-hosted agent back to imperative provisioning (out of TF)

agentydragon:
## Why

The `claude-managed-agents` OpenTofu provider (`modus-agendi/anthropic-claude-managed-agents`, **v1.1.0 latest release + unreleased `main`**) has **no `self_hosted` support**. Its `environment` resource makes `networking` a **required** attribute and always marshals `config.networking`, which the API rejects for `type=self_hosted`:

```text
anthropic api: status=400 type=invalid_request_error
message=config.networking: Extra inputs are not permitted
```

So the self-hosted TF migration (#2673) never applied. For self-hosted, `{type: self_hosted}` is the entire config (egress is ours — the worker pod's mitmproxy). The imperative `ant` path sends exactly that and works (it created the live env, `env_015uqL9WAMSDytQEWWmLG9zF`). Revert self-hosted to imperative; **keep the anthropic-hosted (cloud) agent on TF** — the provider handles `type=cloud` fine.

## Changes

- **Restore** `haku/runtime/managed_agent/self_hosted/{haku.environment,haku.agent,haku.deployment}.yaml` + `provision.sh`, **porting in gmail-labeling** (added across runtimes in #2673) so the imperative config reflects current desired state (MCP server + `mcp_toolset` + vault credential).
- **Rewrite** the dir `README.md` with a "Why this is provisioned imperatively, not Terraform" section.
- **Remove** `tf/gitops/haku-self-hosted-agent` + `cluster/k8s/haku/self-hosted-agent-tf` and its root `kustomization.yaml` entry.
- **Repoint** the `agent-worker` deployment comment at the imperative config.

## Safety

No runtime impact: the live worker already runs on the imperative env (hardcoded `ANTHROPIC_ENVIRONMENT_ID`); the failed `haku-self-hosted-agent` Terraform CR had empty state (`destroyResourcesOnDeletion=false`, `lastAppliedRevision` empty → nothing to destroy). Flux prunes the CR on merge; the empty `haku_self_hosted_agent` tofu-state schema is left as an orphan to drop later.

Follow-up (out of repo): to actually enable gmail-labeling on the **live** self-hosted agent, run the new gmail-labeling vault-credential step + push a new agent version (`ant beta:agents update`) — see the README.


agentydragon:
Added reciprocal cross-surface **SYNC** comments (commit 2683c7e): `model` + the `tana-ro`/`gmail-labeling` MCP servers/toolsets are duplicated between this self-hosted `haku.agent.yaml` and the cloud `tf/gitops/haku-cloud-agent/main.tf`. A true SSOT isn't practical (tofu-controller runner can't read repo-wide files; HCL-vs-YAML split), so each surface now points at its sibling + the shared fields, with an inline note at each `model` line. Runtime C (`haku/runtime/agent`) is env-driven (`HAKU_MODEL`), a separate architecture — not a literal copy.

agentydragon:
Added the **SSOT drift guard** (commit bd96c4c) — upgrading the sync comments from "keep in step by hand" to enforced:

- `haku/base/agent_shared.yaml` — single source of truth for the shared fields (`model`, `tana-ro`/`gmail-labeling` URLs + toolset policies).
- `//haku/base:test_agent_config_ssot` — parses **both** surfaces structurally (self-hosted YAML via PyYAML, cloud HCL via `pygohcl` — no regex) and fails if either drifts from the SSOT. **Verified on RBE**: passes green, and mutating the SSOT model turns it red.
- SYNC comments now point at the SSOT + test.

No live codegen (tofu-controller runner can't read repo-wide files), so values still live in each surface — the test guarantees agreement. Runtime C stays out (env-driven).

PR #2949, opened 2026-07-06
haku/state_template: mirror agent-session Bazel compatibility from haku-state

agentydragon:
## Summary

- Mirrors two generic Bazel-compatibility fixes from `haku-state` into `haku/state_template`, per the template's own sync boundary (`memory/haku-ui.md`'s "Template sync boundary" — generic architecture/method seeds into the template, deployment-specific stays in the live instance).
- Claude Code sessions auto-inject `--config=ai_agent` and `--config=rbe` (assuming every repo touched is ducktape). A fresh haku-state instance built from this template would hit `Config value 'ai_agent'/'rbe' is not defined` the same way haku-state itself did — `.bazelrc` now defines both as local-execution no-ops, matching haku-state's fix.
- Fixes a stale comment in `bazel-ci.yaml`: `eslint` runs as a Bazel target (`//ui/frontend:eslint_test`) via `bazel test //...` in `bazel-ci.yaml`, not in `lint.yaml` (only `ruff` runs there).

## Test plan

- [x] `pre-commit run` on the touched files (yaml/prettier/ducktape hooks all pass)
- [x] Both fixes are verified working as-is in `haku-state` (same repo shape, same Claude Code session environment)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sxebrv3dKVStnxSKEQzatj)_

### DIFF
--- a/haku/state_template/.bazelrc
+++ b/haku/state_template/.bazelrc
@@ -9,6 +9,13 @@ common --enable_bzlmod
 # not here — see tools/ci/trust_egress_ca.sh for the runner's.
 build --spawn_strategy=local
 
+# A Claude Code session's auto-generated buildbuddy.bazelrc unconditionally sets
+# `build --config=rbe`, assuming every repo it touches is ducktape (which defines a real
+# `build:rbe` pointing at BuildBuddy). Define it here as a no-op — genuinely no remote
+# executor is wired below, so this stays exactly as local as the rest of this file — purely so
+# the flag resolves instead of erroring "Config value 'rbe' is not defined" in an agent session.
+build:rbe --spawn_strategy=local
+
 # Strict action env: actions get a fixed PATH instead of inheriting the client's
 # PATH. Keeps builds reproducible across the runner and dev machines.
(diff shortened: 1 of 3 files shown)


haku/state_template/.bazelrc
```diff
@@ -9,6 +9,13 @@ common --enable_bzlmod
 # not here — see tools/ci/trust_egress_ca.sh for the runner's.
 build --spawn_strategy=local
 
+# A Claude Code session's auto-generated buildbuddy.bazelrc unconditionally sets
```
agentydragon: This should really be updated to a to-do to fix our Claude code session tooling in Devinfra/

agentydragon: Good call — pushed `1b894a58`: turned both comments into a `TODO(devinfra-session-tooling)` pointing at `devinfra/claude/claude_hook/main.rs` (the session bazelrc writer that unconditionally assumes every repo is ducktape), rather than just documenting why the shim exists here.

---

PR #3007, opened 2026-07-10
haku/console: full-page past-tool-calls history view (live), panel toggle, deny-with-reason, screenshot target

agentydragon:
## Summary

Adds a separate, full-page view of the whole MCP tool-call ledger to the Haku console, plus several related console improvements. Three commits:

### 1. Full-page "Past tool calls" history view + panel toggle (`e33349e9`)
- New route **`/tool-calls`** — a full page (not a sidebar) listing the whole tool-call ledger with color-coded status (OK / ERROR / DENIED), caller, timestamp, rationale, denial reason, JSON arguments, and collapsible result / id.
- **Routing change**: the console now distinguishes its own pages by URL *path* (`routing.ts`), leaving the hash reserved for mirroring the framed haku-ui route so console navigation never collides with the frame. Production nginx already serves the SPA for any non-asset path; `app.py`'s dev fallback mirrors that so deep links work locally too.
- Backend `GET /api/tool-calls` gains `newest_first=true` so the newest calls survive the query limit (default stays oldest-first for the pending-approval queue).
- Replaced the wrong "Approvals" text toggle with a generic **hamburger icon** carrying a **pulsing callout light** when a tool call awaits approval; added a "Past tool calls" link into the panel. Icons are inline SVG (the `@tabler` barrel OOMs esbuild — see `debug/esbuild_tabler_memory.md`).

### 2. Deny-with-reason + visual screenshot target (`d117c038`)
- Each pending approval card now has an always-visible optional **"Denial reason"** field, so a single Deny click carries a reason straight from the drawer.
- New **`//haku/console/frontend:screenshots`** — a js_test that renders each console surface (history view, approval drawer, access tab) to a PNG in its undeclared outputs for eyeballing. Runs on the RBE worker's display stack via the shared `frontend_visual` puppeteer lib + hermetic Playwright Chromium; a mocked fetch + `<base href>` let the origin-less harness render the data-fetching history view populated. A new `frontend/AGENTS.md` instructs always running it and visually checking the output when editing console visuals.

### 3. Live-update the full history via the shared WS signal (`c958a02c`)
- Extracted the console's tool-call change signal (the `/api/approvals/ws` WebSocket + cross-tab BroadcastChannel) into a `useToolCallEvents(onEvent)` hook. The approval drawer and the `/tool-calls` page both use it, so the full page refetches live on every submit / approve / deny / finish without a reload. The hook only opens the WebSocket from a real http(s) origin, keeping the screenshot harness safe.

## Testing

`bbr test`/`bbr build` all green: `tsc_test`, `bridge_test` (vitest), the production `bundle` (esbuild + tailwind), `test_app`, `test_mcp_approval` (incl. a new `newest_first` ordering test), the new `screenshots` render test, and Python lint (ruff + mypy). All three console surfaces were rendered and visually eyeballed via the new screenshot target.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V5PSCrFPzAn61Yqr2hZE9a)_

### DIFF
--- a/haku/console/frontend/haku_ui_embed.tsx
+++ b/haku/console/frontend/haku_ui_embed.tsx
@@ -27,6 +27,7 @@ import { ShellControls, ShellDrawer, type ShellDrawerTab } from "./console_panel
 import { GEO_PERMISSION_DENIED, GeolocationWatcher, getGeolocation } from "./geolocation.ts";
 import { hasGeolocationGrant, setGeolocationGrant } from "./geolocation_grant.ts";
 import { toastError, toastSuccess } from "./toast.ts";
+import { useToolCallEvents } from "./tool_call_events.ts";
 
 // Haku's own UI service — a separate, Authentik-gated origin running in haku-sandbox —
 // embedded as a sandboxed cross-origin iframe (the whole console is now this frame). The
@@ -52,19 +53,7 @@ export function openExternal(url: string): boolean {
 }
 
 const POPUP_HINT = "Allow pop-ups for this site so the console can open links.";
-const TOOL_APPROVAL_CHANNEL = "haku-console-tool-approvals";
 const MCP_AUTH_CHANNEL = "haku-console-mcp-auth";
-
-type ToolApprovalChannelMessage = { type: "toolApprovalsChanged" };
-
-function isToolApprovalChannelMessage(value: unknown): value is ToolApprovalChannelMessage {
-  return Boolean(
-    value &&
-    typeof value === "object" &&
-    "type" in value &&
-    (value as { type?: unknown }).type === "toolApprovalsChanged"
-  );
-}
 
 // Restore the route the console URL carries into the frame on first mount. The console
 // hash is treated strictly as a validated path, never a URL: the src is always `uiUrl`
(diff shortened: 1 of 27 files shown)


haku/console/frontend/haku_ui_embed.tsx
```diff
@@ -479,12 +448,14 @@ export function HakuUiEmbed({ uiUrl, launchAvailable }: { uiUrl: string; launchA
...
-        refreshToolApprovals(true);
+        refreshToolApprovals();
+        notifyPeers();
       },
       (e: unknown) => {
         setDeciding(approvalId, false);
         toastError("Couldn't deny tool call", e);
-        refreshToolApprovals(true);
+        refreshToolApprovals();
+        notifyPeers();
```
agentydragon: I would expect it would be the server that would send back a notification to peers via websockets. If we do this it would be only an optimization

agentydragon: Agreed — the server (`ToolCallEventHub`) already broadcasts every submit/approve/deny/finish over `/api/approvals/ws` to every connected tab (across replicas via `LISTEN/NOTIFY`), so the client-side cross-tab `BroadcastChannel` was only a latency optimization. Removed it and the `notifyPeers` call sites in fb303e4b; `useToolCallEvents` now just does the initial read plus the server WS subscription.

---

agentydragon: Per your "send that removal as its separate PR" — I've moved the removal out of this PR into its own stacked PR, #3008. So #3007 now keeps the `BroadcastChannel` as a behavior-preserving refactor of the existing approval mechanism, and #3008 removes it on top. (It has to be stacked rather than off `devel` because it depends on the `useToolCallEvents` hook this PR introduces.)

---

PR #3088, opened 2026-07-12
haku/console: drop the x-authentik-* operator-auth header fallback

agentydragon:
Fixes the HIGH-severity authentication bypass documented in #3085 — the `x-authentik-*` fallback is a forward-auth-proxy holdover and is dropped completely.

## The bug

`operator_subject` / `operator_username` (`haku/console/operator_auth.py`) fell back to trusting the client-supplied `x-authentik-uid` / `x-authentik-username` request headers whenever the OIDC session had no authenticated user. Those are the retired forward-auth outpost's headers. With the outpost gone (the `HTTPRoute` points straight at the console Service, no `NetworkPolicy`, no edge header strip), any internet client could forge them to pass `require_operator` / `require_operator_or_static_agent` and then reach + self-approve tool calls that execute against the console's own Google / routine credentials.

## The fix

Operator identity now comes **only** from the app-owned OIDC session:

- `operator_subject` returns the session subject or `None` — no header fallback.
- `operator_username` returns `operator_from_session(request)` — no header fallback.

These were the only two readers of `x-authentik-*` in the console, so the header is no longer trusted anywhere. In the `operator_oidc`-unset dev/test mode there is no `SessionMiddleware`, so the operator-only surface is simply unavailable rather than header-spoofable. Production always sets `operator_oidc`.

## Tests

The old tests used the `x-authentik-*` headers as a stand-in for the operator session (which only worked because of the very fallback being removed). Replaced with a real injected operator session:

- `conftest.py` gains `operator_session_cookie(...)` (mirrors Starlette `SessionMiddleware`'s sign format) and a `make_client(operator=True)` mode that runs the app in the production app-owned-auth path (`SessionMiddleware` + active router guards) and sets the signed session cookie on the client.
- `test_mcp_approval.py` / `test_mcp_server.py` updated to that mode; every `x-authentik-*` header is gone from the tests.
- `test_operator_auth.py` already exercised the real Authentik authorization-code login end-to-end and is unchanged; `test_guards_noop_without_operator_oidc` still covers the unset dev/test mode.

`bbr test //haku/console/...` — 19/19 green (lint + mypy on): BuildBuddy invocation `07e5f4c0-d1bd-4e5f-8069-9b597d12c429`.

## Not in scope (defense-in-depth follow-ups)

The code change alone closes the bypass. Optional hardening the review notes but which this PR doesn't include: an edge `RequestHeaderModifier` strip of `x-authentik-*` on the `HTTPRoute`, and a `NetworkPolicy` on the console matching the sibling services (haku-ui/grocy/scanner).

---
_Generated by [Claude Code](https://claude.ai/code/session_013vaCofRJW8ffN137s7C5Dp)_


haku/console/test_mcp_approval.py
```diff
@@ -912,10 +857,11 @@ def test_full_audit_log_listing_and_secret_redaction(
 def test_postgres_store_runs_alembic_and_persists_typed_ledger(
     make_client, tmp_path: Path, db_url: str, mcp_server_url: str
 ) -> None:
-    with make_client(config_file=_config_file(tmp_path, mcp_server_url), csrf_secret=SecretStr("csrf")) as client:
+    with make_client(
+        config_file=_config_file(tmp_path, mcp_server_url), csrf_secret=SecretStr("csrf"), operator=True
```
agentydragon: shared pytest fixture for config file? maybe make_client as fixture factory that uses it by default? maybe same for csrf secret? maybe make_operator_client shorthand?

agentydragon: Done in 5e1884f7. Added:
- `make_operator_client` (conftest) — `make_client` with `operator=True` baked in.
- `console_config` fixture — the standard two-server config.
- `operator_client` fixture — operator session + `console_config` + CSRF, which most operator tests need.

Standard-config operator tests now just take `operator_client` (no `with make_client(...)` block); bespoke-config ones (operator_oauth / gmail) call `make_operator_client(...)`; the two non-operator standard tests take `console_config`. Green: `bbr` invocation `e832f62d-3134-4174-8f65-1d415d03c486`.

---

